Issue #3, opened 2022-05-24
Question/Bug: Support more than 6 significant digits


ckennerknecht:
Hello,

I am having problems (and maybe found a bug) with exporting large double values to my local QuestDB instance.
I have made some test cases to demonstrate them. 
Any help is appreciated!

## General Question:
Is there a way to increase the significant digits for double values that are sent via ILP,
because per default only 6 significant digits are supported.
I could not find any configuration for this.

--> See test cases 1 and 2

## Problem
When the scientific notation is used (7 digits or more before the decimal point) in the line protocol and it seems like QuestDB can not parse this.
[InfluxDB supports the scientific notation.](https://docs.influxdata.com/influxdb/v1.8/write_protocols/line_protocol_reference/#examples)

--> See test case 3

## Setup
```sql
CREATE TABLE 'test' (
 ts TIMESTAMP,
 sensor_id SYMBOL,
 value double
)
timestamp(ts) PARTITION BY Day;
```
```cpp
 sender
   .table(table_name)
   .symbol(id_name, "sensor_1"_utf8)
   .column(value_name, value)
   .at_now();  
```

### Case 1
```cpp
  double value = 1.1234567;
```
Result in WebUI:
![image](https://user-images.githubusercontent.com/47369419/170052414-de30c8b4-ef01-4c39-82d5-ef72ecec7380.png)

Rounded to only 6 significant digits.

### Case 2
```cpp
 double value = 654321.123;
```
Result in WebUI:
![image](https://user-images.githubusercontent.com/47369419/170053743-b7f2b3d2-c464-438e-8c75-250c337c8230.png)

Only 6 significant digits. Digits after decimal point are ignored.

### Case 3
```cpp
 double value = 7654321.123;
```
Console:
> Could not flush buffered messages: An established connection was aborted by the software in your host machine.

QuestDB Log:
> E i.q.c.l.t.LineTcpConnectionContext [2756] could not parse measurement, INVALID_FIELD_VALUE at 38, line (may be mangled due to partial parsing): 'test,sensor=sensor_1 value=7.65432e+06'

As said, the scientific notation, that is used for 7 digits or more, can not be parsed by QuestDB.

amunra:
Thank you for the detailed bug report!

This is really two bugs.

1. The first one is a limitation in the float-to-string logic used in the library. I'm looking right now to fix this.
2. The second is a bug in QuestDB's parsing logic for ILP: I'll create a new bug report there and cross-reference this one.

[...]

PR #5, opened 2022-05-27
fix: Float to string serialization improvements with better accuracy.

amunra:
Now using the [`dtoa.c` library](http://www.netlib.org/fp/) to convert `double` values to strings.
The library encodes floating points as strings that should round-trip back to doubles accurately.

This change fixes issues #3 (Question/Bug: Support more than 6 significant digits).


src/mem_writer.c
```diff
@@ -139,8 +141,106 @@ void mem_writer_i64(mem_writer* writer, int64_t num)
...
+    char* buf_end = NULL;
+
+    // Returns buffer of integer digits or special value, no '.' chars.
+    char* buf = dtoa(num, mode, ndigits, &decpt, &sign, &buf_end);
+    assert((buf_end != NULL) && (buf_end >= buf));
+
+    const size_t len = buf_end - buf;
+    assert(len != 0);
+
+    if (buf[0] == 'N')
```
eugenels, line 169: NaN/Inf values can be tested with isnan/isinf direcly

src/mem_writer.c
```diff
@@ -139,8 +141,106 @@ void mem_writer_i64(mem_writer* writer, int64_t num)
...
+            mem_writer_str(writer, 9, "-Infinity");
+        else
+            mem_writer_str(writer, 8, "Infinity");
+    }
+    else if (decpt <= 0)
+    {
+        char* dest = NULL;
+        const size_t leading_zeros = (size_t)(-decpt);
+        size_t book_len = 2 + leading_zeros + len;
+        if (sign)
```
eugenels, line 185: Can be converted to
```
        size_t book_len = sign + 2 + leading_zeros + len;
        dest = mem_writer_book(writer, book_len);
        if (sign) *dest++ = '-';
```
and save a few lines of code.

amunra: Good point, almost. Really should be `size_t book_len = !!sign + 2 + leading_zeros + len;` (as sign may not be exatcly 1).

eugenels: You are right.
Anyway dtoa_r set it to 1 as was expected.

amunra:
Abandoning change. Will use a different solution that causes fewer threading constraints.

PR #10, opened 2022-06-26
feat: Optional TLS Support

amunra:
# Whole connection encryption using TLS.

## Feature

There are three TLS options:
* Disabled
* Enabled
* Insecure Skip Verify

The enabled option can either use well-known CAs or you can specify a file path to your own CA.

TLS options are passed in via the `sec_opts` argument before connecting.

## Relationship with Authentication

This change is orthogonal to authentication: Either feature may be turned on or off independently.

See examples for more details.


puzpuzpuz (approved):
Didn't check validity of Rust-specific changes from the language perspective. Otherwise, LGTM

PR #18, opened 2022-07-28
Compile as rust lib

mizinovmv:
use lib with rust program

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 
 [lib]
 name = "questdb_client"
-crate-type = ["cdylib", "staticlib"]
+crate-type = ["lib", "cdylib", "staticlib"]
 
 [dependencies]
 libc = "0.2"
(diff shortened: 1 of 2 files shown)


amunra:
Hi!

Thanks for your PR!

I will accept its goals rather than implementation as I intend to clean up and more cleanly split up the Rust code from the bindings for faster compile times.

I'll be doing some refactoring and updating API docs and upload to crates.io, most likely as `questdb-rs` (`questdb::ingress::Sender` to mirror our Python API, with `ingress` as a module so we may add more features in other modules going forward).

I expect this refactoring/publishing to take me roughly until the end of the week.
Whilst the Rust API will change in looks, it will not in its semantics.

You can follow the latest changes in https://github.com/questdb/c-questdb-client/tree/rust_client.

I would welcome your feedback if you'd like to take a look in the meantime.

amunra:
This is now done, and I hope you find this useful!

Check out the `questdb-rs` crate!
* https://crates.io/crates/questdb-rs
* https://docs.rs/questdb-rs/latest/

Thanks for the nudge 😃 !

PR #19, opened 2022-08-02
feat: Rust API.

amunra:
* Split Rust code into `questdb-rs` and `questdb-rs-ffi` crates.
* The intention is to publish the first crate, but not the latter.
* Added Rust docs.


questdb-rs/src/ingress/mod.rs
```diff
@@ -186,6 +126,10 @@ impl <'a> TableName<'a> {
 
         Ok(Self { name: name })
     }
+
+    pub unsafe fn new_unchecked(name: &'a str) -> Self {
```
eugenels, line 130: This fn signature is a bit misleading. 
Unsafe doesn't make much sense for &str. &str must always be valid UTF-8.

questdb-rs-ffi/src/lib.rs
```diff
@@ -313,10 +314,10 @@ pub struct line_sender_table_name
 }
 
 impl line_sender_table_name {
-    fn as_name<'a>(&self) -> TableName<'a> {
-        let str_name = unsafe { std::str::from_utf8_unchecked(
-            slice::from_raw_parts(self.buf as *const u8, self.len)) };
-        TableName{ name: str_name }
+    unsafe fn as_name<'a>(&self) -> TableName<'a> {
```
eugenels, line 317: This fn may be safe

questdb-rs/src/ingress/mod.rs
amunra: Done.

questdb-rs-ffi/src/lib.rs
amunra: Done.

eugenels (approved):

PR #25, opened 2022-08-24
docs(ilp): add rust auth example

jerrinot:
also export all auth examples in the manifest file
so they can be consumed by docs.


amunra (approved):
Lookin' good!

PR #29, opened 2022-11-07
fix: C++ API improvements.

amunra:
* Fixed linkage issue (some `inline` keywords missing).
* Removed alloc from `line_sender_buffer`'s ctor.

### DIFF
--- a/include/questdb/ilp/line_sender.hpp
+++ b/include/questdb/ilp/line_sender.hpp
@@ -195,7 +195,7 @@ namespace questdb::ilp
          * auto validated = "A UTF-8 encoded string"_utf8;
          * @endcode
          */
-        utf8_view operator "" _utf8(const char* buf, size_t len)
+        inline utf8_view operator "" _utf8(const char* buf, size_t len)
         {
             return utf8_view{buf, len};
         }
@@ -206,7 +206,7 @@ namespace questdb::ilp
          * auto table_name = "events"_tn;
          * @endcode
          */
-        table_name_view operator "" _tn(const char* buf, size_t len)
+        inline table_name_view operator "" _tn(const char* buf, size_t len)
         {
             return table_name_view{buf, len};
         }
@@ -217,7 +217,7 @@ namespace questdb::ilp
          * auto column_name = "events"_cn;
          * @endcode
          */
-        column_name_view operator "" _cn(const char* buf, size_t len)
+        inline column_name_view operator "" _cn(const char* buf, size_t len)
         {
             return column_name_view{buf, len};
         }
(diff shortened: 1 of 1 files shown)


include/questdb/ilp/line_sender.hpp
```diff
@@ -271,11 +271,10 @@ namespace questdb::ilp
         line_sender_buffer(
             size_t init_capacity,
             size_t max_name_len) noexcept
-            : _impl{::line_sender_buffer_with_max_name_len(max_name_len)}
+            : _impl{nullptr}
```
eugenels, line 274: What was the intention to defer buffer allocation?
Comparing buffer API with std::vector:

```
std::vector<int> v;
assert(0 == v.capacity()); // no allocation
std::vector<int> v(5);
assert(5 == v.capacity()); // allocation
```

```
questdb::ilp::line_sender_buffer b;
// expecting
assert(0 == b.capacity()); // no allocation
// but
b.capacity() == 64 * 1024; // no allocation
```

```
questdb::ilp::line_sender_buffer b(5);
// expecting
assert(5 == b.capacity()); // allocation
// but
b.capacity() == 5; // no allocation happened
```

amunra: It's to support concurrent queues.

A use case is to have N threads producing buffers and post them on a "pending" queue on for sending.

It would be a waste to reallocate those sent buffers, so instead we can place them on an "available" queue.

Producer threads can then check for buffers from this "available" queue before allocating new ones.

The API for most queues is:

```
bool q::try_deque(T&)
```

So we need a cheap object to pass in here.
In turn, this means deferring alloc.

amunra: What about returning `0` if `!_impl`?

```
        size_t capacity() const noexcept
        {
            if (_impl)
            {
                return ::line_sender_buffer_capacity(_impl);
            }
            else
            {
                return 0;  // Instead of current impl: return _init_capacity;
            }
        }
```

amunra: I can also rename `init_capacity` with `first_alloc_capacity`, but I don't think it'd make much difference.

eugenels: Well, got it.
How about splitting then a constructor with a default parameter into two:
non-allocating `line_sender_buffer() `and allocating `line_sender_buffer(int capacity) `and
hardcode `init_capacity` (aka `first_alloc_capacity`) as a constant?

amunra: Yes, I could split. The downside is that the first time around when used, by default, users will get a tiny buffer initially.
I really doubt anyone _actually_ cares about buffer sizes in practice when creating objects  so maybe it's just fine. Let me know what you think.

Details:

Splitting it up, the default size will be that of Rust's `String::new()`, which in turn just uses Rust's `Vec` growing logic: https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#295.

This in turn ends up growing by doubling the alloc capacity (https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#379) with a minimum alloc size of 8 bytes (https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#112).

In short, it would be a change in behaviour.. then again, probably no one would care/notice.

amunra: I've changed `capacity()` as so:

```cpp
        size_t capacity() const noexcept
        {
            if (_impl)
                return ::line_sender_buffer_capacity(_impl);
            else
                return 0;
        }
```

I think this addresses most of your concerns without really changing existing default behaviour.

eugenels (approved):

PR #30, opened 2022-11-30
feat: Faster string serialization by escaping in byte space.

amunra:
During linux `perf` profiling of the python client, I've noticed that the `write_escaped_unquoted` function was on the hot path.

profile_graph (1)

This PR improves its performance by around ~31%. 

It does so by using unsafe APIs for manipulating strings as vectors: Character escaping is done as bytes rather than chars.

We also get some extra speed by avoiding a branch in the escaping loop by unsafely setting the vector length and setting its contents rather than calling `.push` repeatedly.

We got an extra bit of performance by inlining two functions that get called repeatedly.

Profiling code:

```python
    def test_string_escaping_10m(self):
        count = 10_000_000
        slist = [f's={i:09}==abc \\' for i in range(count)]
        series = pd.Series(slist, dtype='string[pyarrow]')
        df = pd.DataFrame({
            'col1': series,
            'col2': series,
            'col3': series,
            'col4': series,
            'col5': series,
            'col6': series})
        
        buf = qi.Buffer()

        # Warm up and pre-size buffer
        buf.pandas(df, table_name='tbl1', symbols=True)
        buf.clear()

        # Run
        t0 = time.monotonic()
        buf.pandas(df, table_name='tbl1', symbols=True)
        t1 = time.monotonic()
        print(f'Time: {t1 - t0}, size: {len(buf)}')
```

Perf difference:
* With the old code, three sample runs took: 4.12s, 4.01s, 4.07s (avg: 4.06s).
* With the code in this PR, three sample runs took: 2.74s, 2.79s, 2.83s (avg: 2.78s).

### DIFF
--- a/questdb-rs/src/ingress/mod.rs
+++ b/questdb-rs/src/ingress/mod.rs
@@ -736,6 +736,7 @@ impl Buffer {
         self.state = State::Init;
     }
 
+    #[inline(always)]
     fn check_state(&self, op: Op) -> Result<()> {
         if (self.state as isize & op as isize) > 0 {
             return Ok(());
@@ -748,6 +749,7 @@ impl Buffer {
         Err(error)
     }
 
+    #[inline(always)]
     fn validate_max_name_len(&self, name: &str) -> Result<()> {
         if name.len() > self.max_name_len {
             return Err(error::fmt!(
--- a/questdb-rs/src/ingress/mod.rs
+++ b/questdb-rs/src/ingress/mod.rs
@@ -433,10 +433,15 @@ fn write_escaped_impl<Q, C>(
         unsafe { output_vec.set_len(index + additional) };
         for b in s.bytes() {
             if check_escape_fn(b) {
-                output_vec[index] = b'\\';
+                unsafe {
+                    *output_vec.get_unchecked_mut(index) = b'\\';
+                }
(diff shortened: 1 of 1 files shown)


questdb-rs/src/ingress/mod.rs
```diff
@@ -409,44 +409,51 @@ fn write_escaped_impl<Q, C>(
...
-        for c in s.chars() {
-            if check_escape_fn(c) {
-                output.push('\\');
+        let additional = s.len() + to_escape;
+        output_vec.reserve(additional);
+        let mut index = output_vec.len();
+        unsafe { output_vec.set_len(index + additional) };
+        for b in s.bytes() {
+            if check_escape_fn(b) {
+                output_vec[index] = b'\\';
```
eugenels, line 436: You can go even further: `unsafe { *output_vec.get_unchecked_mut(index) = b'\\'; }`

amunra: Ah! I forgot the index was checked! Thanks!

amunra: That was quite a perf boost too. Now the function is almost 30% faster compared to the pre-PR implementation.

eugenels (approved):

PR #47, opened 2023-11-21
feat: Support for using OS certs in Rust

amunra:
Allow chosing between using bundled `webpki-roots` TLS certificate roots, or use the ones provided by the OS.

### DIFF
--- a/questdb-rs/Cargo.toml
+++ b/questdb-rs/Cargo.toml
@@ -22,7 +22,7 @@ base64ct = { version = "1.6.0", features = ["alloc"] }
 ring = "0.17.5"
 rustls = "0.21.7"
 rustls-pemfile = "1.0.3"
-webpki-roots = "0.25.2"
+webpki-roots = { version = "0.25.2", optional = true }
 ryu = "1.0.15"
 itoa = "1.0.9"
 chrono = { version = "0.4.30", optional = true }
@@ -42,10 +42,13 @@ mio = { version = "0.8.8", features = ["os-poll", "net"] }
 chrono = "0.4.30"
 
 [features]
-default = ["tls-native-certs"]
+default = ["tls-webpki-certs"]
 
-# Use OS-provided root TLS certificates
+# Allow use OS-provided root TLS certificates
 tls-native-certs = ["dep:rustls-native-certs"]
+
+# Allow use of the `webpki-roots` crate to validate TLS certificates.
+tls-webpki-certs = ["dep:webpki-roots"]
 
 # Allow skipping verification of insecure certificates.
 insecure-skip-verify = ["rustls/dangerous_configuration"]
--- a/questdb-rs/Cargo.toml
+++ b/questdb-rs/Cargo.toml
@@ -19,13 +19,14 @@ libc = "0.2"
 socket2 = "0.5.4"
 dns-lookup = "2.0.2"
 base64ct = { version = "1.6.0", features = ["alloc"] }
-ring = "0.16.20"
+ring = "0.17.5"
 rustls = "0.21.7"
 rustls-pemfile = "1.0.3"
 webpki-roots = "0.25.2"
 ryu = "1.0.15"
 itoa = "1.0.9"
 chrono = { version = "0.4.30", optional = true }
+rustls-native-certs = { version = "0.6.3", optional = true }
 
 [target.'cfg(windows)'.dependencies]
 winapi = { version = "0.3.9", features = ["ws2def"] }
@@ -41,6 +42,11 @@ mio = { version = "0.8.8", features = ["os-poll", "net"] }
 chrono = "0.4.30"
 
 [features]
+default = ["tls-native-certs"]
+
+# Use OS-provided root TLS certificates
+tls-native-certs = ["dep:rustls-native-certs"]
+
 # Allow skipping verification of insecure certificates.
 insecure-skip-verify = ["rustls/dangerous_configuration"]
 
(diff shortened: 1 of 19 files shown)


questdb-rs/Cargo.toml
```diff
@@ -41,6 +42,14 @@ mio = { version = "0.8.8", features = ["os-poll", "net"] }
 chrono = "0.4.30"
 
 [features]
+default = ["tls-webpki-certs"]
```
jerrinot, line 45: shouldn't we include also the `tls-native-certs` feature by default?

questdb-rs/src/ingress/mod.rs
```diff
@@ -1337,6 +1348,21 @@ fn add_webpki_roots(root_store: &mut rustls::RootCertStore) {
     }));
 }
 
+#[cfg(feature = "tls-native-certs")]
+fn add_os_roots(root_store: &mut RootCertStore) -> Result<()> {
+    let os_certs = rustls_native_certs::load_native_certs().map_err(|io_err| {
+        error::fmt!(TlsError, "Could not OS native TLS certificates: {}", io_err)
```
jerrinot, line 1354: typo: missing a verb.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1337,6 +1348,21 @@ fn add_webpki_roots(root_store: &mut rustls::RootCertStore) {
...
+    })?;
+    for cert in os_certs {
+        root_store
+            .add(&rustls::Certificate(cert.0))
```
jerrinot, line 1358: `add()` contract says this:
```
    /// Add a single DER-encoded certificate to the store.
    ///
    /// This is suitable for a small set of root certificates that are expected to parse
    /// successfully. For large collections of roots (for example from a system store) it
    /// is expected that some of them might not be valid according to the rules rustls
    /// implements. As long as a relatively limited number of certificates are affected,
    /// this should not be a cause for concern. Use [`RootCertStore::add_parsable_certificates`]
    /// in order to add as many valid roots as possible and to understand how many certificates
    /// have been diagnosed as malformed.
```
This tells me not all certificates from the OS trust store can be added and failing to add a cert is not necessarily a critical error. OTOH: Just silently ignoring it does would lead to hard to troubleshoot issues so that's not great either. WDYT?

jerrinot: thinking out-loud: maybe just logging it at some lower level (DEBUG or similar) is the best thing to do here? It would allow for troubleshooting yet it would not fail clients to start in the case OS store has certificates which rustls dislikes for some reason.

questdb-rs/Cargo.toml
amunra: At the moment it's not really required for existing users.
Is it really worth adding the extra dependency for them?

jerrinot: well, if the default behaviour is to combine both bundled and system trust-store (as discussed earlier) then it's needed anyway, isn't it?

questdb-rs/src/ingress/mod.rs
amunra: Fixed, thanks!

questdb-rs/src/ingress/mod.rs
amunra: This is what `reqwest` (very treaded code) does: https://github.com/seanmonstar/reqwest/blob/6792f697fcdb27c47dcbf7bd05f23368d1d4ac80/src/async_impl/client.rs#L485

amunra: I should probably do the same. Generally, I don't think this code should log: It's just a library.

jerrinot: the reqwest strategy looks sensibe: they fail to start only if they manage to load no certificate.

amunra: I've updated the code now. It does the same, but by calling the `add_parsable_certificates` fn.
No logging on error though.

questdb-rs/Cargo.toml
amunra: There's no default behaviour on the Rust side. The default behaviour is going to be on the Python side.
The FFI includes both.

jerrinot: by thinking: if we consider this to be a good default why not make it the default for everyone? 
and if you care about a slim dependencies then you can always exclude default features. wdyt?

questdb-rs/src/ingress/mod.rs
jerrinot: looks good to me!

questdb-rs/Cargo.toml
jerrinot: ok, fair. I am not that familiar with the Rust culture. so I chose to trust your instinct for the expected behaviour.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1337,6 +1348,28 @@ fn add_webpki_roots(root_store: &mut rustls::RootCertStore) {
...
+            io_err
+        )
+    })?;
+
+    let os_certs: Vec<Vec<u8>> = os_certs.into_iter().map(|cert| cert.0).collect();
+    let (valid_count, invalid_count) = root_store.add_parsable_certificates(&os_certs[..]);
+    if valid_count == 0 && invalid_count > 0 {
+        return Err(error::fmt!(
+            TlsError,
+            "zero valid certificates found in native root store ({} found but were invalid)",
```
jerrinot, line 1366: nit: "zero valid certificates" sounds strange to me.
I would personally use "no valid certificate found .." but up to you, I understand it's subjective.

jerrinot (approved):
good work! feel free to resolve that one nit without any action.

questdb-rs/src/ingress/mod.rs
amunra: Done.

PR #49, opened 2023-11-23
fix: more descriptive error messages

jerrinot:
also: fix a possible panic when public keys are too long


amunra (approved):
Code and tests look great to me!

PR #58, opened 2024-03-18
chore: case-sensitive config string

amunra:


amunra:
> Any docs changes needed to highlight this?

Not at this stage. The config format was agreed to be case-sensitive, but the edited changes were older than that agreement. This simply bring this in line with what other clients are doing.

PR #63, opened 2024-03-22
Improve docs on the ingress module

### LINKED ISSUE #60: Error in Getting Started example
The Getting Started at https://crates.io/crates/questdb-rs has the following example:

> This Rust crate supports a number of optional features.
> For example, if you want to work with ILP/HTTP and work with Chrono timestamps, use:
> `cargo add questdb-rs --features ilp-over-http chrono`

but when I execute `cargo add questdb-rs --features ilp-over-http chrono` I get this error:
> error: feature `ilp-over-http` must be qualified by the dependency it's being activated for, like `questdb-rs/ilp-over-http`, `chrono/ilp-over-http`

I tried to add `chrono` as a dependency first and only then run `cargo add questdb-rs --features ilp-over-http chrono` but I'm still getting the same error.

### LINKED ISSUE #61: Incomplete ingres documentation
[This page](https://docs.rs/questdb-rs/4.0.0/questdb/ingress/index.html) was partially updated to accomodate for ILP@HTTP changes, but it looks incomplete:
1. It says:
    > The ingress module implements QuestDB’s variant of the [InfluxDB Line Protocol](https://questdb.io/docs/reference/api/ilp/overview/) (ILP) over TCP.
  
    This is misleading at best. The client supports both TCP and HTTP transport. The example below this sentence uses the HTTP Transport
2. The paragraph [Connection Security Options](https://docs.rs/questdb-rs/4.0.0/questdb/ingress/index.html#connection-security-options) has a code sample which [links to a URL](https://questdb.io/docs/reference/api/ilp/authenticate) which returns 404:
    image
3. The document is silent on configuration, error-handling, retrying, etc. Chances are it's described [elsewhere](https://docs.rs/questdb-rs/4.0.0/questdb/ingress/struct.Sender.html#method.from_conf), but as a Rust developer, I would expect this document to give me an overview of all important concepts.

mtopolnik:
Closes #61. Closes #60.


questdb-rs/src/ingress/mod.rs
```diff
@@ -26,12 +26,17 @@
 //!
 //! The `ingress` module implements QuestDB's variant of the
 //! [InfluxDB Line Protocol](https://questdb.io/docs/reference/api/ilp/overview/)
-//! (ILP) over TCP.
+//! (ILP) over HTTP and TCP. We recommend using HTTP, and consider TCP as a
+//! legacy protocol variant.
```
amunra, line 30: I think it's too early to call TCP legacy.
Rather, I think it should state why HTTP.

Something like "We recommend using the new ILP/HTTP protocol since it provides better error feedback and fine transaction control.".

questdb-rs/src/ingress/mod.rs
```diff
@@ -26,12 +26,17 @@
...
-//!   * Connect to QuestDB by creating a [`Sender`] object, usually via [`Sender::from_conf`].
-//!   * Populate a [`Buffer`] with one or more rows of data.
-//!   * Send the buffer via the Sender's [`flush`](Sender::flush) method.
+//!   * Make sure you have the dependency with the HTTP feature enabled:
+//!
+//! ```cargo add questdb-rs --features ilp-over-http```
+//!
+//!   * Use [`Sender::from_conf()`] to get the [`Sender`] object
+//!   * Populate a [`Buffer`] with one or more rows of data
+//!   * Send the buffer using [`sender.flush()`](Sender::flush)
```
amunra, line 39: I think it's OK to revert this doc change and simply make `ilp-over-http` default as you earlier suggested.

questdb-rs/src/ingress/mod.rs
```diff
@@ -43,87 +48,147 @@
...
+//!    let mut buf = Buffer::new();
+//!    buf
 //!        .table("sensors")?
 //!        .symbol("id", "toronto1")?
 //!        .column_f64("temperature", 20.0)?
 //!        .column_i64("humidity", 50)?
 //!        .at(TimestampNanos::now())?;
-//!    sender.flush(&mut buffer)?;
+//!    sender.flush(&mut buf)?;
 //!    Ok(())
```
amunra, line 59: Revert, unnecessary change.

questdb-rs/src/ingress/mod.rs
```diff
@@ -26,12 +26,17 @@
 //!
```
amunra, line 26: Good changes.

The docs here are getting non-trivial in size.

How about moving them with a

```
#![doc = include_str!("mod.md")]
```

And then move these docs to `questdb/src/ingress/mod.md`.

questdb-rs/src/ingress/mod.rs
```diff
@@ -43,87 +48,147 @@
...
 //!        .column_i64("humidity", 50)?
 //!        .at(TimestampNanos::now())?;
-//!    sender.flush(&mut buffer)?;
+//!    sender.flush(&mut buf)?;
 //!    Ok(())
 //! }
 //! ```
 //!
-//! # Flushing
+//! # Unified APIs: configuration string and builder API
```
amunra, line 63: I'm not sure "Unified APIs" means much. Maybe just "Configuration Parameters"?
The content itself is great.

questdb-rs/src/ingress/mod.rs
```diff
@@ -621,10 +679,10 @@ impl BufferState {
 /// use questdb::ingress::{Buffer, TimestampMicros, TimestampNanos};
 ///
 /// # fn main() -> Result<()> {
-/// let mut buffer = Buffer::new();
+/// let mut buf = Buffer::new();
```
amunra, line 682: buf -> buffer.

Examples consistently use the latter.

questdb-rs/src/ingress/mod.rs
```diff
@@ -675,26 +732,27 @@ impl BufferState {
...
 /// To insert a NULL value, skip the symbol or column for that row.
 ///
 /// # Recovering from validation errors
 ///
-/// If you want to recover from potential validation errors, you can use the
-/// [`set_marker`](Buffer::set_marker) method to track a last known good state,
-/// append as many rows or parts of rows as you like and then call
-/// [`clear_marker`](Buffer::clear_marker) on success.
+/// If you want to recover from potential validation errors, call
+/// [`buf.set_marker()`](Buffer::set_marker) to track the last known good state,
```
amunra, line 750: buf -> buffer

amunra (changes requested):
Substantial improvements.
Some minor tweaks recommended.
Note that the C, C++ and FFI doc strings probably need updating as well: I've been trying to keep them in sync.

amunra:
Clarification:

I had written feedback about reverting `buffer` -> `buf` changes.
I don't particularly care _which_ variable name is used, so long as it's consistent with the rest of the examples in `questdb-rs/examples` as well.

mtopolnik:
I had a specific reason to use `buf` instead of `buffer` --  making the construct `sender.flush(&mut buffer)` a bit shorter, as I was going to use it in sentences whenever referring to `flush`. But in the end, I use it just once, so I'll revert to `buffer`.

questdb-rs/src/ingress/mod.rs
mtopolnik: This heading covers just the single paragraph below it, which explains that there's the Builder struct whose API is identical to the configuration string.

mtopolnik:
`cargo test --no-default-features` doesn't compile on either `main` or this branch.

amunra:
Sorry. Correct. Need to enable at least webpki-roots.

mtopolnik:
`test_tls_ca_webpki_and_os_roots` is failing with

```
LineProtocolException: [-1] table: 77836915f4494765acbd56ec18f6dd7e, column: a; cast error from protocol type: TAG to column type: STRING
```

My guess is that there's something wrong with the isolation between tests. I suppose this test should create its own table, but it seems to already exist and there's a type clash on column `a`.

mtopolnik:
Working on the docs, I noticed that `line_sender.hpp` doesn't declare `flush_and_keep_with_flags`, so it seems you can't ensure a transactional flush from C++.

questdb-rs/README.md
```diff
@@ -1,15 +1,17 @@
...
-Official Rust client for [QuestDB](https://questdb.io/), an open-source SQL database designed to process time-series data, faster.
+Official Rust client for [QuestDB](https://questdb.io/), an open-source SQL
+database designed to process time-series data, faster.
 
-The client library is designed for fast ingestion of data into QuestDB via the InfluxDB Line Protocol (ILP).
+The client library is designed for fast ingestion of data into QuestDB via the
+InfluxDB Line Protocol (ILP).
 
 * [QuestDB Database docs](https://questdb.io/docs/)
 * [ILP docs](https://questdb.io/docs/reference/api/ilp/overview/)
```
jerrinot, line 10: nit: consider changing 'ILP docs' to something like: "InfluxDB Line Protocol Documentation". 
In general, I would the ILP acronym when possible.

jerrinot: image

tris0laris: agree - I would keep ILP to a minimum and replace with InfluxDB Line Protocol, people do not know what ILP means

questdb-rs/src/ingress/mod.md
```diff
@@ -0,0 +1,232 @@
+# Fast Ingestion of Data into QuestDB
+
+The `ingress` module implements QuestDB's variant of the
+[InfluxDB Line Protocol](https://questdb.io/docs/reference/api/ilp/overview/)
+(ILP) over HTTP and TCP. We recommend using HTTP because it has a number of
```
jerrinot, line 5: Do you want to bother users with HTTP vs TCP from the very start, before showing them any code at all? why? Could we be opinionated in the begging and show this "decision point" only later? that's what I did with Java client: https://questdb.io/docs/clients/java_ilp/#transport-selection

questdb-rs/src/ingress/mod.md
```diff
@@ -0,0 +1,232 @@
...
+features that TCP lacks, such as receiving error messages from the server and
+fine-grained transaction control.
+
+To get started:
```
jerrinot, line 9: this is great! 3 lines to give you the basic usage pattern! I might steal this for Java. I have something similar, but more detailed. I think yours is better for a quick overview of the expected usage pattern.

questdb-rs/src/ingress/mod.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Sequential Coupling in the Buffer API
+
+The fluent API of [`Buffer`] has sequential coupling: there's a certain order in
+which you are expected to call the methods. For example, you must write the
+symbols before the columns, and you must terminate each row by calling either
+[`at`](Buffer::at) or [`at_now`](Buffer::at_now). Refer to the [`Buffer`] doc
+for the full rules and a flowchart.
+
+## Flushing
```
jerrinot, line 179: consider moving this up in the docs? IMHO grasping flushing (and error handling) is perhaps the most important concept user need to understand. even more so given the Rust client does not auto-flush.

jerrinot (changes requested):
overall it's a pretty great improvement! 
I see only two bigger things:
1. consider deemphasizing the TCP transport. Move it from the Getting Started chapter (=the material for absolute beginners, it's OK to be opinionated there)
2. explain error handling in the Rust Ingres docs. I think that's pretty important for production usage.

questdb-rs/src/ingress/mod.md
mtopolnik: I'm moving it up and rephrasing a bit. But keep in mind that this issue is way more obvious in Rust, which doesn't use the `Sender` object to add data. It uses two separate objects: `Buffer` and `Sender`, so it jumps out at you that you must pass the buffer to sender to get any action.

jerrinot (approved):

jerrinot (approved):

PR #71, opened 2024-04-26
chore: Add RELEASING.md, minor update to mod.md

mtopolnik:
A paragraph was in the wrong place in mod.md (rust doc of the ingress module).

### DIFF
--- a/doc/RELEASING.md
+++ b/doc/RELEASING.md
@@ -1,0 +1,89 @@
+# How to publish a new release of the Rust ILP Client library
+
+## 1. Ensure you're fully in sync with the remote repo
+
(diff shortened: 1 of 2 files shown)


doc/RELEASING.md
```diff
@@ -0,0 +1,89 @@
+# How to publish a new release of the Rust ILP Client library
```
amunra, line 1: Nit:
Title implies this is only for Rust, when it's for the whole of c-questdb-client.

amunra (changes requested):
Minor tweak.

amunra (approved):

PR #77, opened 2024-07-09
fix: rustls version conflict when using questdb-rs

jerrinot:
lock ureq to 2.9
a workaround for https://github.com/algesten/ureq/issues/765


mtopolnik (approved):

corrosion/.github/workflows/gh-pages.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
jerrinot, line 27: what's the reason for this update? is it compatible?

corrosion/.github/actions/setup_test/action.yaml
```diff
@@ -138,10 +138,10 @@ runs:
         arch: ${{ steps.arch_flags.outputs.msvc }}
       if: ${{ 'msvc' == steps.determine_abi.outputs.abi }}
     - name: Install CMake
-      uses: corrosion-rs/install-cmake@v2
+      uses: lukka/get-cmake@519de0c7b4812477d74976b2523a9417f552d126
```
jerrinot, line 141: why this change?

mtopolnik: These are changes inside the `corrosion` subtree, which I just copied in full from corrosion github

amunra (approved):
Looks great!
As an extra bonus I've seen the new updated dependences have fewer transitive dependencies too. Nice!

PR #88, opened 2025-02-28
Implement C++ line_sender::flush_and_keep_with_flags()

dekinet:
Implements the (missing) C++ line_sender::flush_and_keep_with_flags() method.
I do not know how to use/interface to rust, so have relied heavily on the existing line_sender::flush_and_keep() method.


dekinet:
Hi, the failing checks seem entirely unrelated to the change I made?

Issue #89, opened 2025-03-11
Groundwork to prepare for arrays


amunra:
We will soon be introducing an nd-array type in QuestDB and in the ILP protocol.

* The new array fields will be sent as binary.
* The existing `Buffer` is internally based on a mutable `String` type in Rust: This needs to be modified to internally use a `Vec<u8>` instead.
* The existing methods that expose the buffer's accumulated string also need to be removed and replaced with equivalent methods that expose bytes. C, C++ and Python APIs need to be adjusted accordingly.
* We also need to update our `ureq` dependency to 3.x. This will mean reworking some of the TLS logic since the new library more clearly decouples these details.

The resulting PR will thus effectively introduce a minor but nevertheless breaking change in the API.

PR #94, opened 2025-03-18
BREAKING CHANGE: buffers exposed as binary, as a UTF-8 string

kafka1991:
# Overview

This change introduces a minor breaking API change.

In preparation for follow-up work, the ILP buffer will change from always internally serializing to a UTF-8 string to serializing to binary data.

This PR changes the API to allow this, without changing any of the functionality or wire format.

# Detail

The APIs that exposes the buffer in Rust (`.as_str()`), C (`line_sender_buffer_peek`) and C++ (`buffer::peek()`) have been changed to return a binary buffer.

In Rust specifically, the `.as_str()` method has been renamed to `.as_bytes()`.

This PR closes https://github.com/questdb/c-questdb-client/issues/89

Internally this change swaps `Buffer`'s backing from a `String` to a `Vec<u8>`.

### DIFF
--- a/cpp_test/test_line_sender.cpp
+++ b/cpp_test/test_line_sender.cpp
@@ -40,11 +40,18 @@ using namespace std::string_literals;
 using namespace questdb::ingress::literals;
 
 template <typename F>
-    class on_scope_exit
+class on_scope_exit
 {
 public:
-    explicit on_scope_exit(F&& f) : _f{std::move(f)} {}
-    ~on_scope_exit() { _f(); }
+    explicit on_scope_exit(F&& f)
+        : _f{std::move(f)}
+    {
+    }
+    ~on_scope_exit()
+    {
+        _f();
+    }
+
 private:
     F _f;
 };
@@ -53,26 +60,23 @@ TEST_CASE("line_sender c api basics")
 {
     questdb::ingress::test::mock_server server;
     ::line_sender_error* err = nullptr;
-    on_scope_exit error_free_guard{[&]{
-            if (err)
-                ::line_sender_error_free(err);
-        }};
+    on_scope_exit error_free_guard{[&] {
+        if (err)
+            ::line_sender_error_free(err);
+    }};
     ::line_sender_utf8 host = {0, nullptr};
     CHECK(::line_sender_utf8_init(&host, 9, "localhost", &err));
-    ::line_sender_opts* opts = ::line_sender_opts_new(
-        ::line_sender_protocol_tcp,
-        host,
-        server.port());
+    ::line_sender_opts* opts =
+        ::line_sender_opts_new(::line_sender_protocol_tcp, host, server.port());
     CHECK_NE(opts, nullptr);
     ::line_sender* sender = ::line_sender_build(opts, &err);
     line_sender_opts_free(opts);
     CHECK_NE(sender, nullptr);
     CHECK_FALSE(::line_sender_must_close(sender));
-    on_scope_exit sender_close_guard{[&]
-        {
-            ::line_sender_close(sender);
-            sender = nullptr;
-        }};
+    on_scope_exit sender_close_guard{[&] {
+        ::line_sender_close(sender);
+        sender = nullptr;
+    }};
     server.accept();
     CHECK(server.recv() == 0);
     ::line_sender_table_name table_name{0, nullptr};
@@ -100,7 +104,8 @@ TEST_CASE("line_sender c api basics")
     CHECK(server.msgs().front() == "test,t1=v1 f1=0.5 10000000\n");
 }
 
-TEST_CASE("Opts service API tests") {
+TEST_CASE("Opts service API tests")
+{
     // We just check these compile and link.
 
     line_sender_utf8 host = QDB_UTF8_LITERAL("localhost");
@@ -110,28 +115,24 @@ TEST_CASE("Opts service API tests") {
         ::line_sender_protocol_tcp,
         ::line_sender_protocol_tcps,
         ::line_sender_protocol_http,
-        ::line_sender_protocol_https
-    };
+        ::line_sender_protocol_https};
 
-    for (size_t index = 0; index < sizeof(protocols) / sizeof(::line_sender_protocol); ++index)
+    for (size_t index = 0;
+         index < sizeof(protocols) / sizeof(::line_sender_protocol);
+         ++index)
     {
         auto proto = protocols[index];
-        ::line_sender_opts* opts1 = ::line_sender_opts_new_service(
-            proto,
-            host,
-            port);
+        ::line_sender_opts* opts1 =
+            ::line_sender_opts_new_service(proto, host, port);
         ::line_sender_opts_free(opts1);
     }
-
 }
 
 TEST_CASE("line_sender c++ connect disconnect")
 {
     questdb::ingress::test::mock_server server;
     questdb::ingress::line_sender sender{
-        questdb::ingress::protocol::tcp,
-        "localhost",
-        server.port()};
+        questdb::ingress::protocol::tcp, "localhost", server.port()};
     CHECK_FALSE(sender.must_close());
     server.accept();
     CHECK(server.recv() == 0);
@@ -149,8 +150,7 @@ TEST_CASE("line_sender c++ api basics")
     CHECK(server.recv() == 0);
 
     questdb::ingress::line_sender_buffer buffer;
-    buffer
-        .table("test")
+    buffer.table("test")
         .symbol("t1", "v1")
         .symbol("t2", "")
         .column("f1", 0.5)
@@ -167,8 +167,7 @@ TEST_CASE("test multiple lines")
 {
     questdb::ingress::test::mock_server server;
     std::string conf_str =
-        "tcp::addr=localhost:" +
-        std::to_string(server.port()) + ";";
+        "tcp::addr=localhost:" + std::to_string(server.port()) + ";";
     questdb::ingress::line_sender sender =
         questdb::ingress::line_sender::from_conf(conf_str);
     CHECK_FALSE(sender.must_close());
@@ -177,8 +176,7 @@ TEST_CASE("test multiple lines")
 
     const auto table_name = "metric1"_tn;
     questdb::ingress::line_sender_buffer buffer;
-    buffer
-        .table(table_name)
+    buffer.table(table_name)
         .symbol("t1"_cn, "val1"_utf8)
         .symbol("t2"_cn, "val2"_utf8)
         .column("f1"_cn, true)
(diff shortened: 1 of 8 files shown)


include/questdb/ingress/line_sender.hpp
```diff
@@ -424,20 +424,16 @@ namespace questdb::ingress
...
-            {
-                size_t len = 0;
-                const char* buf = ::line_sender_buffer_peek(_impl, &len);
-                return {buf, len};
-            }
-            else
-            {
-                return {};
-            }
+        std::string_view peek_as_bytes() const noexcept {
```
amunra, line 430: Returned type should be `std::span<const std::byte>`.

amunra: Keep the old name `peek()`.

questdb-rs-ffi/src/lib.rs
```diff
@@ -672,20 +672,29 @@ pub unsafe extern "C" fn line_sender_buffer_transactional(
     buffer.transactional()
 }
 
-/// Get a string representation of the contents of the buffer.
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct line_sender_buffer_view {
+    len: size_t,
+    buf: *const c_char,
```
amunra, line 679: `buf: *const u8`, not `c_char`

include/questdb/ingress/line_sender.h
```diff
@@ -360,18 +368,15 @@ LINESENDER_API
 bool line_sender_buffer_transactional(const line_sender_buffer* buffer);
 
 /**
- * Get a string representation of the contents of the buffer.
+ * Get a read-only view into the buffer's bytes content..
```
amunra, line 371: `content..` -> `contents.`

include/questdb/ingress/line_sender.h
```diff
@@ -171,6 +171,14 @@ line_sender_utf8 line_sender_utf8_assert(size_t len, const char* buf);
 #define QDB_UTF8_LITERAL(literal)                                              \
     line_sender_utf8_assert(sizeof(literal) - 1, (literal))
 
+/**
+ * Non-owning view of sender buffer.
```
amunra, line 175: Add an extra note about safety here:
"Modifying the buffer will invalidate the borrowed buffer"

include/questdb/ingress/line_sender.h
```diff
@@ -171,6 +171,14 @@ line_sender_utf8 line_sender_utf8_assert(size_t len, const char* buf);
...
+ */
+typedef struct line_sender_buffer_view {
+  size_t len;
+  const char *buf;
```
amunra, line 179:
```
const char* uint8_t
          ^ ^______ more appropriate type for binary.
          |________ style, associate the pointer to the type, not the variable name.
```

questdb-rs/src/tests/sender.rs
```diff
@@ -71,13 +71,14 @@ fn test_basics() -> TestResult {
...
         ts_nanos_num
     );
-    assert_eq!(buffer.as_str(), exp);
+    let exp_byte = exp.as_bytes();
+    assert_eq!(buffer.as_bytes(), exp_byte);
     assert_eq!(buffer.len(), exp.len());
     sender.flush(&mut buffer)?;
     assert_eq!(buffer.len(), 0);
-    assert_eq!(buffer.as_str(), "");
+    assert_eq!(buffer.as_bytes(), "".as_bytes());
```
amunra, line 79: assert_eq!(buffer.as_bytes(), b"");

questdb-rs/src/tests/sender.rs
```diff
@@ -305,7 +307,7 @@ fn test_chrono_timestamp() -> TestResult {
     buffer.table(tbl_name)?.column_ts("a", ts)?.at(ts)?;
 
     let exp = "tbl_name a=1000000t 1000000000\n";
```
amunra, line 309: This can just be:
```
let exp = b"tbl_name a=1000000t 1000000000\n";
assert_eq!(buffer.as_bytes(), exp);
```

questdb-rs/src/tests/sender.rs
```diff
@@ -305,7 +307,7 @@ fn test_chrono_timestamp() -> TestResult {
...
-    assert_eq!(buffer.as_str(), exp);
+    assert_eq!(buffer.as_bytes(), exp.as_bytes());
```
amunra, line 310: This can just be:
```
let exp = b"tbl_name a=1000000t 1000000000\n";
assert_eq!(buffer.as_bytes(), exp);
```

system_test/questdb_line_sender.py
```diff
@@ -104,6 +104,10 @@ class c_line_sender_utf8(ctypes.Structure):
 class c_line_sender_table_name(ctypes.Structure):
     _fields_ = [("len", c_size_t),
                 ("buf", c_char_p)]
+class line_sender_buffer_view(ctypes.Structure):
+    _fields_ = [("len", c_size_t),
+                ("buf", c_char_p)]
```
amunra, line 109: uint8

amunra (changes requested):
Nearly there!
Some data type inaccuracies to correct and _very minor_ code cleanup required.

cpp_test/test_line_sender.cpp
```diff
@@ -382,46 +383,57 @@ TEST_CASE("Validation of bad chars in key names.")
...
-    CHECK(buffer1.peek() == "buffer3");
-    CHECK(buffer3.peek() == "");
-    CHECK(buffer3.size() == 0);
-    CHECK(buffer3.capacity() == 0);
-    CHECK(buffer3.peek() == "");
+#if __cplusplus >= 202002L
+bool operator==(const std::span<const std::byte> &lhs, const char *rhs) {
+  return lhs.size() == std::strlen(rhs) &&
+         std::memcmp(lhs.data(), rhs, lhs.size()) == 0;
+}
```
amunra, line 390: A few things here to change:
* spans are small objects, they should be passed by value.
* Code style! Align `&` and `*` to the right. New line for braces. 4 spaces indent. If an IDE is messing things up either disable the style rules or tweak them.
* The `const char*` paired with `strlen` will ignore any nulls, meaning you will not be able to assert arrays. Instead take the size in as a template argument.
* memcmp is a bit of a C way of doing things. Probably best just convert the string to a span and compare the ranges.

In other words:
```
template <std::size_t N>
bool operator==(std::span<const std::byte> lhs, const char (&rhs)[N]) 
{
    constexpr std::size_t bytelen = N - 1; // Exclude null terminator
    const std::span<const std::byte> rhs_span{reinterpret_cast<const std::byte*>(rhs), bytelen};
    return lhs.size() == bytelen && std::ranges::equal(lhs, rhs_span);
}
```

cpp_test/test_line_sender.cpp
```diff
@@ -382,46 +383,57 @@ TEST_CASE("Validation of bad chars in key names.")
...
 
+bool operator==(const std::span<const std::byte> &lhs,
+                const std::string &rhs) noexcept {
+  return lhs.size() == rhs.size() &&
+         std::memcmp(lhs.data(), rhs.data(), lhs.size()) == 0;
```
amunra, line 395: Ditto:
* Fix style
* Convert the string to a span and compare that.

include/questdb/ingress/line_sender.hpp
```diff
@@ -95,6 +99,23 @@ namespace questdb::ingress
         https,
     };
 
+    struct line_sender_buffer_bytes {
```
amunra, line 102: Move this to `#else` branch of `#if __cplusplus >= 202002L`.
No point defining it for C++20 if it's unused.

Also consider inheriting from the C type and adding the accessor methods for ptr and type that mimick the `std::span<const std::byte>` API.

include/questdb/ingress/line_sender.hpp
```diff
@@ -423,21 +444,22 @@ namespace questdb::ingress
                 return 0;
         }
 
+#if __cplusplus >= 202002L
+        using buffer_view_t = std::span<const std::byte>;
```
amunra, line 448: Rename `buffer_view_t` to `buffer_view`.

include/questdb/ingress/line_sender.hpp
```diff
@@ -423,21 +444,22 @@ namespace questdb::ingress
...
+#else
+        using buffer_view_t = line_sender_buffer_bytes;
```
amunra, line 450: Move impl here, remove typedef.

questdb-rs/src/ingress/mod.rs
```diff
@@ -251,12 +251,11 @@ impl From<Infallible> for Error {
     }
 }
 
-fn write_escaped_impl<Q, C>(check_escape_fn: C, quoting_fn: Q, output: &mut String, s: &str)
+fn write_escaped_impl<Q, C>(check_escape_fn: C, quoting_fn: Q, output_vec: &mut Vec<u8>, s: &str)
```
amunra, line 254: For consistency with the other functions, you might want to rename `output_vec` to `output`.

CMakeLists.txt
```diff
@@ -16,7 +16,12 @@ set(CMAKE_C_STANDARD 11)
 set(CMAKE_C_STANDARD_REQUIRED ON)
 set(CMAKE_C_EXTENSIONS OFF)
 
-set(CMAKE_CXX_STANDARD 17)
+set(MINIMUM_REQUIRED_CXX_STANDARD 17)
+
+if((NOT CMAKE_CXX_STANDARD) OR (CMAKE_CXX_STANDARD LESS MINIMUM_REQUIRED_CXX_STANDARD))
+    set(CMAKE_CXX_STANDARD ${MINIMUM_REQUIRED_CXX_STANDARD})
+endif()
```
amunra, line 23: Question:
Are we sure that our old code never worked with C++14?
If so, we probably want to continue supporting it too.

include/questdb/ingress/line_sender.hpp
```diff
@@ -333,6 +318,163 @@ namespace questdb::ingress
...
+        /// @return Const pointer to the data (may be nullptr if empty()).
+        constexpr const std::byte* data() const noexcept { return buf; }
+
+        /// @brief Gets the number of bytes in the view.
+        /// @return Size in bytes.
+        constexpr size_t size() const noexcept { return len; }
+
+        /// @brief Checks if the view contains no bytes.
+        /// @return true if size() == 0.
+        constexpr bool empty() const noexcept { return len == 0; }
```
amunra, line 362: Keep `data()`, `size()`, `empty()` and maybe the equality operator.

Remove the rest: It's too much.

We don't need to backport std-span, just provide a minimal API that's forward compatible with span.

The comments can also be cleaned up: Remove the section markers, too much as well.

Small is beautiful and easier to read :-)

include/questdb/ingress/line_sender.hpp
```diff
@@ -445,22 +587,35 @@ namespace questdb::ingress
...
+         */
+        std::span<const std::byte> peek() const noexcept
+        {
+            if (_impl)
+            {
+                auto view = ::line_sender_buffer_peek(_impl);
+                return {reinterpret_cast<const std::byte*>(view.buf), view.len};
+            }
+            return {};
+        }
```
amunra: Introduce a typedef in C++20 called `buffer_view` and remove the macro logic.

cpp_test/test_line_sender.cpp
amunra: Introducing `.clang-format` is _real_ fix to it.

Thanks!

amunra (approved):
Looking good!

PR #96, opened 2025-03-20
chore(rust): swapped crypto backend to aws-lc and updated crate dependencies for ureq, rustls, webpki-roots and others

kafka1991:
# Overview

* Upgraded HTTP library from `ureq` 2.9 to 3.0.10.
    * The ureq version is partially pinned due to its non-strict SemVer compliance.
* Switched default cryptographic backend from `ring` to `aws-lc-rs`, ring is still available via a feature flag.
* Updated the `rustls` and `webpki-roots` dependencies to the latest versions.
* Upgraded to the latest questdb-confstr-rs, removing the need for a trailing semicolon in config strings.

This PR closes https://github.com/questdb/c-questdb-client/issues/89


questdb-rs/src/ingress/mod.rs
```diff
@@ -2364,6 +2372,126 @@ impl SenderBuilder {
     }
 }
 
+pub struct TlsConnector {
```
amunra, line 2375: Move to `http`, or it will be compiled in even if HTTP is disabled in the feature flags.

questdb-rs/src/tests/http.rs
```diff
@@ -237,7 +237,7 @@ fn test_no_connection() -> TestResult {
     let err = res.unwrap_err();
     assert_eq!(err.code(), ErrorCode::SocketError);
     assert!(err.msg().starts_with(
-        "Could not flush buffer: http://127.0.0.1:1/write?precision=n: Connection Failed"
+        "Could not flush buffer: Io error: Connection refused"
```
amunra, line 240: Not good to lose this detail in the error message: Letting a user know what they were connecting to could help them debug a configuration issue.

amunra:
TODO: Must check that NO_DELAY is set for new HTTP impl.

amunra:
> > TODO: Must check that NO_DELAY is set for new HTTP impl.
> 
> I think we have set the `NO_DELAY=true` flag in new ureq. image

Yes, indeed. I'd seen it - no worries!
I just wanted to be sure that i'd bet picked up properly through all the chaning layers and in the newer ureq3 codebase.

amunra (approved):

PR #98, opened 2025-04-01
introduce ndarr support

kafka1991:
introduce ndarr support

**TODO Lists**
- [x]  Line protocol version tests in c, c++ and system tests (rust client done)
- [x]  Rust client's support for common built-in array and vector APIs (for ease of use, less 3 dims)
- [x] add more system array tests
- [x] Python client support

### DIFF
--- a/questdb-rs-ffi/src/ndarr.rs
+++ b/questdb-rs-ffi/src/ndarr.rs
@@ -746,7 +746,7 @@ mod tests {
 
     #[test]
     fn test_stride_minus_non_contiguous_strides_array() -> TestResult {
-        let test_data = vec![
+        let test_data = [
             1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 11.0, 12.0,
         ];
         let shape = [2usize, 2];


questdb-rs/src/ingress/mod.rs
```diff
@@ -993,6 +995,121 @@ impl Buffer {
...
+        }
+
+        let index = self.output.len();
+        let reserve_size = reserve_size.checked_add(index).ok_or(error::fmt!(
+            ArrayViewError,
+            "Array total elem size overflow"
+        ))?;
+        self.output
+            .try_reserve(reserve_size)
+            .map_err(|_| error::fmt!(BufferOutOfMemory, "Buffer out of memory"))?;
```
amunra, line 1105: That's not how reserve works: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.reserve

You pass in the argument for _additional_ required size, _not the total_.

Also `try_reserve` is completely pointless here since we can end up with large memory pressure already: A regular `reserve` should do.

questdb-rs/src/ingress/mod.rs
```diff
@@ -993,6 +995,121 @@ impl Buffer {
...
+        unsafe { self.output.set_len(reserve_size) }
+
+        // ndarr data
+        view.write_row_major_buf(&mut self.output[index..]);
```
amunra, line 1109: Security!
The memory would most likely not be `null`, I'll need to think about this in more depth as it could be an attack vector in the application.
Consider calling https://doc.rust-lang.org/std/vec/struct.Vec.html#method.resize instead.

...alternatively we change the API so that we can check if the array exposes a buffer that we can directly read instead, if not then we fall back to doing this instead.

Come to think about it, we could have three code paths for better efficiency:

* A code path for borrowing a readable buffer.
* A code path for taking ownership of a buffer (see `Bytes::from_owner`)
* A code path to write elements one by one.

questdb-rs/src/ingress/ndarr.rs
```diff
@@ -0,0 +1,133 @@
+pub(crate) const MAX_DIMS: usize = 32;
```
amunra, line 1: I'd move this top-level in `questb::ingress` and make it public, then add a doc to it. It might be useful to callers.

questdb-rs/src/ingress/mod.rs
amunra: Ok.. there's certainly no attack vector that I can think of, but it is easy for an API to write too little data and corrupt things or to read uninitialized memory.
This bit I think we ought to do better.

I suggest:
```rust
// NB: I haven't run any of this, just trying to provide an alternative idea.
let writable = unsafe {
    std::slice::from_raw_parts_mut(
        self.output.ptr_mut().add(self.output.len()),
        array_buf_size)
};
let cursor = Cursor::new(writable);
view.write_row_major(&mut cursor);
if cursor.position() != array_buf_size {
    return Err(... tell the user that their array write op was the wrong size ...);
}

// Update last so in case of errors we haven't grown the buffer.
unsafe { self.output.set_len(new_len) };
```

It should be logically very similar, one extra check and if I'm not wrong (I could be) the cursor should not perform an extra copy, but this last detail should probably be triple-checked by reading the standard lib impl.

amunra: I forgot to mention.. the trait method would then have a different signature:
```
fn write_row_major<W: std::io::Write>(&self, dest: &mut W) -> std::io::Result<()>;
```

questdb-rs/src/error.rs
```diff
@@ -48,6 +48,15 @@ pub enum ErrorCode {
...
     ConfigError,
+
+    /// Array has too many dims. Currently, only arrays with a maximum [`crate::ingress::ndarr::MAX_DIMS`] dimensions are supported.
+    ArrayHasTooManyDims,
+
+    /// Array view internal error.
+    ArrayViewError,
+
+    /// Buffer outOfMemory.
+    BufferOutOfMemory,
```
amunra, line 59: Remove

questdb-rs/src/ingress/ndarr.rs
```diff
@@ -0,0 +1,133 @@
...
+    /// Returns the corresponding data type classification.
+    ///
+    /// This enables runtime type identification while maintaining
+    /// compile-time type safety.
+    fn elem_type() -> ElemDataType;
+}
+
+#[repr(u8)]
+#[derive(Debug, PartialEq)]
+pub enum ElemDataType {
```
amunra, line 36: Where does this come from?
Is there a spec document that maps the supported datatypes?

If there's a format document link to it, otherwise let's write a spec for it and link to it.
This enum here is just too confusing and too error-prone otherwise.

Each entry should also be explicit about its value, e.g. `Boolean = 1`, `Byte = 2` .. etc.

Additionally, why is there an `Undefined` entry?

amunra: PS, do we really intend to support all of these datatypes as binary?

questdb-rs/src/ingress/mod.rs
```diff
@@ -24,8 +24,8 @@
 
 #![doc = include_str!("mod.md")]
 
+pub use self::ndarr::*;
```
amunra, line 27: The `*` risks leaking too many types.
Instead just pick the ones you explicitly are sure you want in the public API.

questdb-rs/src/tests/mod.rs
```diff
@@ -29,6 +29,9 @@ mod http;
 mod mock;
 mod sender;
 
+#[cfg(feature = "ndarray")]
+mod ndarr;
```
amunra, line 33: There should be tests for the trait directly as well, even without `ndarray` enabled.

This is to be sure that we never tie the `ndarray` feature to the ability to write arrays.

The tests that check the `NdArrayView` trait should also check what happens when the trait does nothing or less data.

amunra: Another test to write is if the trait gives back the wrong number of dimensions.

include/questdb/ingress/line_sender.hpp
```diff
@@ -624,6 +626,39 @@ class line_sender_buffer
...
+     *
+     * @note Data is stored contiguously in row-major (C-style) order.
+     *       Example: shape [2,3] expects 6 elements ordered as:
+     *       [a11, a12, a13, a21, a22, a23]
+     */
+    template <typename T, size_t N>
+    line_sender_buffer& column(
+        column_name_view name,
+        const std::vector<uint32_t>& shape,
+        const std::array<T, N>& data)
```
amunra, line 645: We need to come up with something that's a bit more idiomatic C++.
Let's park the C++ API for now until we have everything else nailed down.

questdb-rs/src/ingress/ndarr.rs
amunra: Then this list should only contain double :-)

questdb-rs/src/ingress/ndarr.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn dim(&self, index: usize) -> Option<usize> {
+        let len = self.ndim();
+        if index < len {
+            Some(self.len_of(Axis(index)))
+        } else {
+            None
+        }
+    }
+
+    fn write_row_major_buf(&self, buf: &mut [u8]) {
```
amunra, line 114: The implementation here is incomplete.

It should handle cases where the strides are non-row-major standard.
The current code will serialize the same buffer for both array A and its transpose A'.

amunra: Worth having a bunch of independent tests that just test the serialization of arrays to buffers for different arrays of different shapes and strides.

questdb-rs/src/tests/mock.rs
```diff
@@ -519,20 +519,8 @@ impl MockServer {
...
-                let tail = index + 1;
-                let msg = std::str::from_utf8(&accum[head..tail]).unwrap();
-                self.msgs.push(msg.to_owned());
-                head = tail;
-                received_count += 1;
-            }
-        }
-        Ok(received_count)
+        self.msgs = accum;
+        Ok(self.msgs.len())
```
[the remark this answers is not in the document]
amunra: Yeah, but ignoring the problem is not a solution :-)

This makes the tests frail and prone to occasionally fail.

questdb-rs/src/ingress/ndarr.rs
```diff
@@ -0,0 +1,754 @@
...
+
+pub(crate) fn get_and_check_array_bytes_size<A: NdArrayView<T>, T>(
+    array: &A,
+) -> Result<usize, Error>
+where
+    T: ArrayElement,
+{
+    (0..array.ndim())
+        .try_fold(std::mem::size_of::<T>(), |acc, i| Ok(acc * array.dim(i)?))
+        .and_then(|p| match p <= MAX_ARRAY_BUFFER_SIZE {
```
amunra, line 135: Match over bool is just an `if` statement.
I also think this code would probably be easier to read if it were iterative.

```rust
let mut size = std::mem::size_of::<T>();
for dim_index in 0..array.ndim() {
    size = size * array.dim(dim_index)?
    if size > MAX_ARRAY_BUFFER_SIZE {
        return Err(...);
    }
}
Ok(size)
```

questdb-rs/src/ingress/mod.rs
```diff
@@ -464,6 +485,8 @@ impl BufferState {
     }
 }
 
+pub trait Buffer1 {}
```
amunra, line 488: Remove

questdb-rs/src/ingress/mod.rs
```diff
@@ -1822,6 +2052,19 @@ impl SenderBuilder {
                 "retry_timeout" => {
                     builder.retry_timeout(Duration::from_millis(parse_conf_value(key, val)?))?
                 }
+
+                #[cfg(feature = "ilp-over-http")]
+                "disable_line_protocol_validation" => {
```
amunra, line 2057: Let's rename this to something better like "autodetect_protocol" where `true` is default.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1011,6 +1069,175 @@ impl Buffer {
...
+
+        // ndarr data
+        ndarr::write_array_data(view, writeable, array_buf_size)?;
+        unsafe { self.output.set_len(array_buf_size + index) }
+        Ok(self)
+    }
+
+    #[allow(private_bounds)]
+    #[cfg(feature = "benchmark")]
+    pub fn column_arr_use_raw_buffer<'a, N, T, D>(&mut self, name: N, view: &T) -> Result<&mut Self>
```
amunra, line 1198: Remove and remove benchmark.

questdb-rs/src/ingress/mod.rs
amunra: This way if a user wants to override they just specify:
```protocol_version=2``` and that's it.

File: questdb-rs/src/ingress/mod.rs [hunk 15]
```diff
@@ -566,6 +593,8 @@ impl Buffer {
             state: BufferState::new(),
             marker: None,
             max_name_len: 127,
+            f64serializer: f64_binary_series,
+            version: LineProtocolVersion::V2,
```
amunra, line 597: No, this needs to be an argument.
We _can't_ default to V2 just yet. It would break things too much.

Rather I think we can have a new utility:

```
let sender = Sender::from_conf(...);
let buffer = sender.new_buffer();  // version auto-populated.
```

I think it would be worth updating examples with this.

questdb-rs/src/ingress/mod.rs
```diff
@@ -582,6 +611,36 @@ impl Buffer {
         buf
     }
 
+    pub fn with_line_proto_version(mut self, version: LineProtocolVersion) -> Result<Self> {
```
amunra, line 614: Remove.
This should be folded into `::new()` as required argument.

questdb-rs/src/ingress/mod.rs
```diff
@@ -582,6 +611,36 @@ impl Buffer {
...
+        }
+        self.f64serializer = match version {
+            LineProtocolVersion::V1 => f64_text_series,
+            LineProtocolVersion::V2 => f64_binary_series,
+        };
+        self.version = version;
+        Ok(self)
+    }
+
+    pub fn set_line_proto_version(&mut self, version: LineProtocolVersion) -> Result<&mut Self> {
```
amunra, line 629: Remove. There should be just one obvious way to do this.

questdb-rs/src/ingress/mod.rs
amunra: Ensure that `Buffer::new()` documents and cross-references `sender.new_buffer()` and vice-versa.

amunra: NB. The protocol should be added to `max_name_len` too.

In actual fact, `server.new_buffer()` should actually call `Buffer::with_max_name_len` under the covers, since this comes from config too.

questdb-rs/src/ingress/mod.rs
```diff
@@ -582,6 +611,36 @@ impl Buffer {
...
+        if self.state.op_case != OpCase::Init {
+            return Err(error::fmt!(
+                LineProtocolVersionError,
+                "Line protocol version must be set before adding any data."
+            ));
+        }
+        self.f64serializer = match version {
+            LineProtocolVersion::V1 => f64_text_series,
+            LineProtocolVersion::V2 => f64_binary_series,
+        };
```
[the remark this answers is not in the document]
amunra: You _could_ bias the V2 with `std::hint::likely`.
Inlining the f64 serialization functions is probably a good idea too.

questdb-rs/src/ingress/mod.rs
```diff
@@ -953,8 +1012,7 @@ impl Buffer {
         Error: From<N::Error>,
     {
         self.write_column_key(name)?;
-        let mut ser = F64Serializer::new(value);
-        self.output.extend_from_slice(ser.as_str().as_bytes());
+        (self.f64serializer)(&mut self.output, value);
```
amunra, line 1015:
```if
if likely(!matches!(version, LineProtocolVersion::V1)) {
    binary
} else {
    text
}
```

questdb-rs/src/ingress/mod.rs
```diff
@@ -55,6 +57,25 @@ use ring::{
     signature::{EcdsaKeyPair, ECDSA_P256_SHA256_FIXED_SIGNING},
 };
 
+/// Defines the maximum allowed dimensions for array data in binary serialization protocols.
+pub const MAX_ARRAY_DIMS: usize = 32;
+
+/// Line Protocol Version supported by current client.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum LineProtocolVersion {
```
amunra, line 65: Rename to `ProtocolVersion` to match what I think is a better config setting, I.e. `protocol_version=1|2|auto`.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1011,6 +1069,175 @@ impl Buffer {
...
+    /// # Examples
+    ///
+    /// Basic usage with direct dimension specification:
+    ///
+    /// ```
+    /// # #[cfg(feature = "ndarray")]
+    /// # {
+    /// # use questdb::Result;
+    /// # use questdb::ingress::Buffer;
+    /// # use ndarray::array;
```
amunra, line 1086: Don't comment the `use ndarray::array` line here. Should be shown as part of the example.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1011,6 +1069,175 @@ impl Buffer {
         Ok(self)
     }
 
+    /// Record a multidimensional array value for the given column.
+    ///
+    /// Supports arrays with up to [`MAX_ARRAY_DIMS`] dimensions. The array elements must
+    /// implement [`ArrayElement`] trait which provides type-to-[`ElemDataType`] mapping.
```
amunra, line 1075: ElemDataType has been renamed now.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1011,6 +1069,175 @@ impl Buffer {
...
+    /// # fn main() -> Result<()> {
+    /// # let mut buffer = Buffer::new();
+    /// # buffer.table("x1")?;
+    /// // Record a 3D array of f64 values
+    /// let array_3d = Array3::from_elem((2, 3, 4), 42f64);
+    /// let col_name = ColumnName::new("col1")?;
+    /// buffer.column_arr(col_name, &array_3d.view())?;
+    /// # Ok(())
+    /// # }
+    /// # }
```
amunra, line 1116: I'd remove this example and show one with slices and another with vectors.

You can use some headings here.

Run
```
cargo doc --open
```

To preview.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1199,6 +1426,9 @@ pub struct Sender {
     handler: ProtocolHandler,
     connected: bool,
     max_buf_size: usize,
+    default_line_protocol_version: LineProtocolVersion,
+    #[cfg(feature = "ilp-over-http")]
+    supported_line_protocol_versions: Option<Vec<LineProtocolVersion>>,
```
amunra, line 1431: Probably worth commenting why it's an `Option<Vec<..>>` and not just a `Vec<..>`.

I'm assuming that's because it's `None` if no probing was done.
I'm just afraid neither of us will rememember in 6 months time.

questdb-rs/src/ingress/mod.rs
```diff
@@ -2092,6 +2335,24 @@ impl SenderBuilder {
...
 
+    #[cfg(feature = "ilp-over-http")]
+    /// Disables automatic line protocol version validation for ILP-over-HTTP.
+    ///
+    /// - When set to `"off"`: Skips the initial server version handshake and disables protocol validation.
+    /// - When set to `"on"`: Keeps default validation behavior (recommended).
+    ///
+    /// Please ensure client's default version ([`LINE_PROTOCOL_VERSION_V2`]) or
+    /// explicitly set protocol version exactly matches server expectation.
+    pub fn disable_line_protocol_validation(mut self) -> Result<Self> {
```
amunra, line 2346: Replace with `pub fn protocol_version(mut self, vers: ProtocolVersion)`, overriding the default when called.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1199,6 +1426,9 @@ pub struct Sender {
     handler: ProtocolHandler,
     connected: bool,
     max_buf_size: usize,
+    default_line_protocol_version: LineProtocolVersion,
```
amunra, line 1429: Replace with `protocol_version: Option<ProtocolVersion>` where `None` means "auto".

questdb-rs/src/ingress/http.rs
```diff
@@ -383,3 +405,175 @@ pub(super) fn http_send_with_retries(
 
     retry_http_send(state, buf, request_timeout, retry_timeout, last_rep)
 }
+
+pub(super) fn get_line_protocol_version(
+    state: &HttpHandlerState,
+    settings_url: &str,
+) -> Result<(Option<Vec<LineProtocolVersion>>, LineProtocolVersion), Error> {
```
amunra, line 412: Document return type: I would have expected a single `Vec<ProtocolVersion>`.

amunra: Also ensure that there's tests for the following scenarios:
* The server returns 404.
  * Default to V1
* The server returns, but the field is missing from the JSON.
  * Default to V1
* The server returns only version 1 (if the v2 protocol gets toggled via config in the server).
* The server returns [1, 2].
  * Client uses V2.
* The server returns [2].
  * Client uses V2. 
* The server returns [MAX_SUPPORTED_VERSION + 1].
  * The client errors, since there's no compatible protocol.
  
For these tests, ensure create a buffer from `server.new_buffer()`, add a simple with just a timestamp and an f64 and flush it, checking the serialized behaviour.

questdb-rs/Cargo.toml
```diff
@@ -55,6 +56,8 @@ mio = { version = "1", features = ["os-poll", "net"] }
 chrono = "0.4.31"
 tempfile = "3"
 webpki-roots = "0.26.8"
+criterion = "0.5"
+rstest = "0.25.0"
```
amunra, line 60: I'd prefer to avoid this extra dependency. I think the tests can be refactored into functions that are called for V1 and V2 variants.

questdb-rs/src/ingress/mod.rs
kafka1991: Currently, `protocol_version=auto` cannot replace the `disable_line_protocol_validation` configuration. Regardless of the value configured for `protocol_version`, the sender will request the server to verify whether the client's configuration is within the server's support range. The role of `disable_line_protocol_validation` is to explicitly turn it off if the user does not want the verification process(a way to trun off the round trip for permance.).

PR #101, opened 2025-05-07
chore: allow external `Buffer` flushability checking

ischeinkman:
Added a `Buffer::check_can_flush` method for checking if a `Buffer` can be flushed with `Sender::flush`. 

## Usecase

In our application we're operating in an `async` setting with a custom `async` HTTP library & writer. At the moment we need to isolate the QuestDB sending logic into its own set of threads because it doesn't expose enough information to be able to flush the `Buffer` ourselves. However, by including this `check_can_flush` call we can now recreate the entirety of `Sender::flush_impl` using our custom library and async system.

### DIFF
--- a/questdb-rs/src/ingress/mod.rs
+++ b/questdb-rs/src/ingress/mod.rs
@@ -1172,6 +1172,31 @@ impl Buffer {
         self.state.row_count += 1;
         Ok(())
     }
+
+    /// Checks if this buffer is ready to be flushed to a sender via one of the
+    /// [`Sender::flush`] functions. An [`Ok`] value indicates that the buffer
+    /// is ready to be written to using the given [`Protocol`] and transaction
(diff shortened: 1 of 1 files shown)


ideoma (approved):

amunra:
Fair :-)

What HTTP library are you using, out of curiosity?

amunra:
I've merged this PR, but note that we're adding array support soon which will also introduce new minor breaking changes to the API. Some of these have already been introduced in the `main` branch.

As such I'll delay releasing version 5.0.0 until those changes are ready.
For now, you should be able to use the `questdb-rs` dependency tied to the specific commit:

```toml
[dependencies]
questdb-rs = { git = "https://github.com/questdb/c-questdb-client", rev = "abc1234def5678" }
```

PR #105, opened 2025-06-03
Stability changes for version 5.0.0

amunra:
# Overview

This PR is about introducing stability changes targeting the release of version 5.0.0.

# Task list
- [x] @javier: Updated front page doc and examples that are consistent across all clients.
- [ ] JSON-driven array tests
- [x] Reword doc sections around protocol version, make it clear that selecting the protocol
  is only suggested for TCP and we advise "auto" for HTTP.
- [x] Add a test for `max_line_len` from `/settings` (see point above).
- [x]  Introduce a new simpler array C API that doesn't require strides:
  `line_sender_buffer_column_f64_arr_row_major` (we will use this from Python).
- [x] Investigate if the two APIs that take strides can be optimised further.
  You mentioned that these are slower than the equivalent Rust `ndarray` APIs.
- [x] Consolidate array error codes into a single `ArrayError` code. These should be consolidated:
    - `ArrayHasTooManyDims`
    - `ArrayViewError`
    - `ArrayWriteToBufferError`
- [x] Introduce a lower-level C++ API that just takes pointers and lengths for
  data and strides.
- [x] Introduce an overloaded C++ API that just takes an array data pointer and
  length, and assumes default strides (calling `_f64_arr_row_major`).
- [x] Fix warnings when generating docs with `(cd questdb-rs && cargo doc --open)`
  E.g.: warning: unresolved link to `Sender::new_buffer_with_max_name_len`.
- [x] Add the C++ API change from PR https://github.com/questdb/c-questdb-client/pull/88
- [x] Update and document (front page + examples + api docs) which is the first
  version of QuestDB supports arrays.
  My guess is that it is going to be 8.4.0, but we need to confirm this.
  (I've also added this todo to the Python client TODO list.)
- [x] Improved C++ API ergonomics:
  - Lower-level API that is similar to the C API
  - Higher-level API that takes a vectors and spans in C++20.
- [ ] Once the next QuestDB release is cut, we should revert the CI hacks that test
  against the `nd_arr` branch. We should also retain a CI test run that runs
  against an older version of QuestDB (e.g. 8.3.3) to ensure that we don't
  introduce regressions against protocol version 1.
- [x] Revise `MAX_ARRAY_BUFFER_SIZE`: lower it!

# Tandem PR
See related `py-questdb-client` PR: https://github.com/questdb/py-questdb-client/pull/108


include/questdb/ingress/line_sender.h
```diff
@@ -504,21 +496,46 @@ bool line_sender_buffer_column_str(
...
+ * @param[out] err_out Set to an error object on failure (if non-NULL).
+ * @return true on success, false on error.
+ */
+LINESENDER_API
+bool line_sender_buffer_column_f64_arr_c_major(
+    line_sender_buffer* buffer,
+    line_sender_column_name name,
+    size_t rank,
+    const uintptr_t* shape,
+    const uint8_t* data_buffer,
```
amunra, line 520: In this and other array APIs the buffer should be typed.
In other words, `const double* data_buffer`.
Same for C++.

PR #106, opened 2025-06-03
Improve docs for the array column

mtopolnik:


include/questdb/ingress/line_sender.h
```diff
@@ -499,18 +499,19 @@ bool line_sender_buffer_column_str(
     line_sender_error** err_out);
 
 /**
- * Record a multidimensional array of double for the given column.
+ * Record a multidimensional array of `double` values for the given column.
```
amunra, line 502: I'd avoid the backticks, since this is doxygen and not markdown.

If you really wanted monospace it's `@c double`, but it's unnecessary.

PR #134, opened 2026-02-28
feat(ingress): add `_opt` builder methods to Buffer for Option<T>

### LINKED ISSUE #133: Ergonomic optional columns: Add _opt methods to Buffer
### Is your feature request related to a problem? Please describe.
When using the `Buffer` API to insert rows, the fluent builder pattern is incredibly ergonomic. However, when working with real-world data structs that contain nullable fields (represented as `Option<T>` in Rust), developers are forced to break the method chain to use `if let` statements. 

Because QuestDB treats missing columns as `NULL`, we have to conditionally call the column methods.

**Example of the current pain point:**
```rust
let mut buffer = Buffer::new(ProtocolVersion::V3);

// We must break the chain to handle optional values
let mut row = buffer.table("sensors")?
      .symbol("location", "factory-1")?;

if let Some(temp) = sensor.temperature {
    row.column_f64("temperature", temp)?;
}

if let Some(hum) = sensor.humidity {
    row.column_f64("humidity", hum)?;
}

row.at_now()?;
```

### Describe the solution you'd like

I propose adding `_opt` variants for all column assignment methods directly to the Buffer struct (e.g., `column_f64_opt`, `column_str_opt`, `symbol_opt`, `column_dec_opt`, etc.).

If the provided value is Some(v), the method behaves exactly like its standard counterpart. If the value is None, it simply returns Ok(self) as a no-op, allowing the builder chain to continue unbroken.

The resulting developer experience:
```rust
let mut buffer = Buffer::new(ProtocolVersion::V3);

buffer
    .table("sensors")?
    .symbol("location", "factory-1")?
    .column_f64_opt("temperature", sensor.temperature)? // Option<f64>
    .column_f64_opt("humidity", sensor.humidity)?       // Option<f64>
    .at_now()?;
```

### Implementation Details
The new methods would reuse the exact same generic constraints currently used in the Buffer API (e.g., N: TryInto<ColumnName<'a>>) to ensure zero performance degradation and maintain compatibility with pre-validated names.

Example signature for strings:
```rust
pub fn column_str_opt<'a, N, S>(&mut self, name: N, value: Option<S>) -> crate::Result<&mut Self>
where
    N: TryInto<ColumnName<'a>>,
    S: AsRef<str>,
    Error: From<N::Error>,
{
    if let Some(v) = value {
        self.column_str(name, v)
    } else {
        Ok(self)
    }
}
```

### Additional context

I have already built this locally using a custom extension trait and it massively cleaned up my ingestion code. I think this would be highly beneficial for the broader community.

If the team is open to this API addition, I would be more than happy to open a PR adding these methods!

aaaronme:
Closes #133

This PR introduces native support for `Option<T>` in the `Buffer` builder API to drastically improve the developer experience when ingesting real-world data with nullable fields. 

### The Problem:
Previously, handling optional data forced users to break the ergonomic method chain:
```rust
buffer.table("trades")?.symbol("symbol", "BTC-USD")?;

if let Some(fee) = trade.maker_fee {
    buffer.column_f64("maker_fee", fee)?;
}
if let Some(order_id) = trade.order_id.as_deref() {
    buffer.column_str("order_id", order_id)?;
}

buffer.at_now()?;
```

### The Solution:
This PR adds `_opt` variants for all column types (`column_str_opt`, `column_f64_opt`, `column_arr_opt`, etc.). If the value is Some, it behaves exactly like the standard method. If None, it is a no-op, skipping the column perfectly in line with ILP standards.

```rust
buffer
    .table("trades")?
    .symbol("symbol", "BTC-USD")?
    .column_f64_opt("maker_fee", trade.maker_fee)?
    .column_str_opt("order_id", trade.order_id.as_deref())?
    .at_now()?;
```
    
### Changes made
Added `_opt` methods directly to `impl Buffer` using the exact same zero-cost generic constraints as their standard counterparts.

Added a new section to the module-level documentation (`# Usage Considerations`) explaining how to handle `NULL`s and how to correctly pass ownership of heap-allocated `Option` types (e.g., using `.as_deref()`).

Documented every new method using intra-doc links to avoid duplicating complex API constraints.

### Testing
As requested in the issue, a rigorous integration test suite was added in `tests/buffer_opt.rs`:

`test_buffer_opt_some_matches_standard`: Uses `ProtocolVersion::V3` to construct one buffer with the standard methods, and one buffer with the `_opt` methods wrapped in Some. Validates that the resulting binary byte arrays are 100% identical (covering Strings, Numbers, Decimals, and Arrays).

`test_buffer_opt_none_skips_columns`: Asserts that passing `None` results in valid ILP syntax where the columns are entirely omitted from the UTF-8 output.

## Summary by CodeRabbit

* **New Features**
  * Added optional-parameter buffer APIs so fields wrapped in Option<T> can be conditionally omitted while preserving fluent chaining.

* **Documentation**
  * Added a guide on representing NULLs and examples showing conditional column writing with optional parameters.

* **Tests**
  * Added tests ensuring optional-parameter behavior matches non-optional behavior for present values and correctly omits columns for None.

### DIFF
--- a/questdb-rs/src/ingress/buffer.rs
+++ b/questdb-rs/src/ingress/buffer.rs
@@ -771,6 +771,45 @@ impl Buffer {
         Ok(self)
     }
 
+    /// Record a symbol for the given column if the value is `Some`.
+    /// If the value is `None`, this is a no-op and the column is skipped.
+    ///
+    /// This is a convenience wrapper around [`Self::symbol`]. Make sure you
+    /// record all symbol columns before any other column type.
+    ///
+    /// **Tip:** If you have an `Option<String>`, use `.as_deref()` to pass it
+    /// without consuming the original string.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use questdb::Result;
+    /// # use questdb::ingress::{Buffer, SenderBuilder};
+    /// # fn main() -> Result<()> {
+    /// # let mut sender = SenderBuilder::from_conf("https::addr=localhost:9000;")?.build()?;
+    /// # let mut buffer = sender.new_buffer();
+    /// # buffer.table("x")?;
+    /// let val: Option<&str> = Some("value");
+    /// buffer.symbol_opt("col_name", val)?;
+    ///
+    /// let no_val: Option<&str> = None;
+    /// buffer.symbol_opt("skipped_col", no_val)?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn symbol_opt<'a, N, S>(&mut self, name: N, value: Option<S>) -> crate::Result<&mut Self>
+    where
+        N: TryInto<ColumnName<'a>>,
+        S: AsRef<str>,
+        Error: From<N::Error>,
+    {
+        if let Some(v) = value {
+            self.symbol(name, v)
+        } else {
+            Ok(self)
+        }
+    }
+
     fn write_column_key<'a, N>(&mut self, name: N) -> crate::Result<&mut Self>
     where
         N: TryInto<ColumnName<'a>>,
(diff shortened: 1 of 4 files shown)


aaaronme:
@amunra Failed check bc of unrelated code

amunra:
@jerrinot and/or @mtopolnik might be able to help.

aaaronme:
@amunra I added a small commit here to fix the issue, just clippy & fmt got updated after the new rust version. if it is all ok i think we are ready for a review. Version in question: `cargo 1.95.0`

jerrinot:
hello @aaaronme, thanks for this PR, this looks like a useful contribution. Increased API surface is a good price to pay to enable idiomatic Rust code.

jerrinot (approved):

jerrinot:
@aaaronme there is one thing you could do: a test to verify that every `column_foo()` has a `column_foo_opt()` twin. 
I can think of a simple pattern matching on a source code, but perhaps you can come up with a better way. up to you.

aaaronme:
@jerrinot Thanks for the approval 🙌  I hope to be able to contribute more in the future :)

I'm thinking we pull in the `buffer.rs` source as a string (`include_str!`) and walk it: for each `pub fn`, look at its header up to the body's `{` and check whether it carries the `N: TryInto<ColumnName<'a>>` bound. That bound is the real "this is a column-writing method" signal. It catches `symbol`, every `column_*`, and any future addition like `uuid` that follows the same pattern, regardless of naming. Then for each one we just assert a sibling with `_opt` exists in the same scan. No regex, no new deps, ~25 lines.

If you want I can add it. 
Should I commit it to this branch or open a new PR?

jerrinot:
@aaaronme no problem at all, thanks for your contribution!
Given this PR is already merged then a new PR is the cleanest. Keep the test simple please, I think even just a regexp would do.

PR #138, opened 2026-03-12
feat: add multi-url failover for HTTP/HTTPS senders

nwoolmer:
## Summary
- Add round-robin address rotation on retriable errors for HTTP/HTTPS protocols
- Multiple addresses via repeated `addr` keys in config strings or `address()` API
- Per-address `ureq::Agent` instances to avoid TLS TargetHost conflicts
- IPv6 bracket notation support (`[::1]:9000`) with bare IPv6 rejection
- Retriable: 421, 500, 502, 503, 504, 507, 509, 523, 524, 529, 599, plus connection refused/reset/aborted IO errors
- C/C++ FFI: new `line_sender_opts_address()` function

### Key design decisions
- Time-bounded retries (not count-bounded) with exponential backoff (10ms initial, 2x, 1000ms cap, ±5ms jitter)
- `Arc<rustls::ClientConfig>` shared across endpoints (cheap atomic clone)
- `rotate()` happens after sleep/timeout check, before the retry request — `current_index` always points to the endpoint that produced the last response

## Test plan
- [x] 232 Rust tests passed (config parsing, IPv6, rotation, failover, 502, 421, connection-refused, settings negotiation, non-retriable, parameterized status codes)
- [x] 47 doc-tests passed

## Summary by CodeRabbit

* **New Features**
  * Multi-address failover with round‑robin distribution across HTTP/HTTPS endpoints
  * Support for IPv6 addresses using bracket notation
  * New public API to configure additional failover addresses

* **Improvements**
  * Expanded retriable error handling and failover behavior
  * Updated QuestDB target version to 9.3.0

* **Tests**
  * Extensive tests for multi-address parsing, failover and rotation scenarios


questdb-rs/src/ingress/sender/http.rs
```diff
@@ -525,15 +563,111 @@ fn retry_http_get(
...
     if !need_retry || retry_timeout.is_zero() {
         return last_rep;
     }
 
-    retry_http_get(state, url, request_timeout, retry_timeout, last_rep)
+    retry_http_get(state, request_timeout, retry_timeout, last_rep)
+}
+
+#[cfg(test)]
+mod tests {
```
nwoolmer: tests can be moved out of here

questdb-rs-ffi/Cargo.toml
```diff
@@ -10,7 +10,7 @@ crate-type = ["cdylib", "staticlib"]
 
 [dependencies]
 libc = "0.2"
-questdb-confstr-ffi = { version = "0.1.1", optional = true }
+questdb-confstr-ffi = { path = "/home/nick/repos/questdb-confstr-rs/questdb-confstr-ffi", version = "0.2.0", optional = true }
```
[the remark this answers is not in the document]
nwoolmer: Need to merge the confstr PR first

questdb-rs-ffi/src/lib.rs
```diff
@@ -1471,6 +1471,22 @@ pub unsafe extern "C" fn line_sender_opts_bind_interface(
...
+/// Multiple addresses are only supported for HTTP/HTTPS protocols.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn line_sender_opts_address(
+    opts: *mut line_sender_opts,
+    host: line_sender_utf8,
+    port: line_sender_utf8,
+    err_out: *mut *mut line_sender_error,
+) -> bool {
+    unsafe { upd_opts!(opts, err_out, address, host.as_str(), port.as_str()) }
+}
```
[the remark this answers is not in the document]
nwoolmer: The `upd_opts\!` macro dummy-builder-on-error pattern is pre-existing across all FFI option setters (not introduced by this PR). Tracking it as a separate issue.

questdb-rs/src/ingress/sender/http.rs
nwoolmer: Done in [`1693714`](https://github.com/questdb/c-questdb-client/commit/1693714) — moved all 6 rotation unit tests to `questdb-rs/src/tests/http.rs`.

questdb-rs/src/ingress/mod.rs
```diff
@@ -685,6 +706,42 @@ impl SenderBuilder {
...
+        if self.protocol.is_tcpx() {
+            return Err(error::fmt!(
+                ConfigError,
+                "Multiple addresses are only supported for HTTP/HTTPS protocols."
+            ));
+        }
+        let host = host.into();
+        let port: Port = port.into();
+        self.addresses.push((host, port.0));
+        Ok(self)
```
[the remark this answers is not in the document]
nwoolmer: Fixed in [`1693714`](https://github.com/questdb/c-questdb-client/commit/1693714). `address()` now rejects empty host and empty port with `ConfigError`. Added two tests: `multi_url_address_rejects_empty_host` and `multi_url_address_rejects_empty_port`.

questdb-rs/src/ingress/mod.rs
```diff
@@ -1121,24 +1181,37 @@ impl SenderBuilder {
...
+                        .timeout_connect(Some(*http_config.request_timeout.deref()))
+                        .http_status_as_error(false);
+                    let agent = ureq::Agent::with_parts(
+                        agent_builder.build(),
+                        connector,
+                        ureq::unversioned::resolver::DefaultResolver::default(),
+                    );
+                    let url = format!("{proto}://{host}:{port}/write");
+                    let settings_url = format!("{proto}://{host}:{port}/settings");
+                    endpoints.push(HttpEndpoint {
```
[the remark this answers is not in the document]
nwoolmer: Fixed in [`1693714`](https://github.com/questdb/c-questdb-client/commit/1693714). IPv6 hosts (detected by `contains(':')`) are now wrapped in brackets: `format\!("{proto}://[{host}]:{port}/write")`.

questdb-rs-ffi/src/lib.rs
```diff
@@ -1435,7 +1435,8 @@ pub unsafe extern "C" fn line_sender_opts_new(
     host: line_sender_utf8,
     port: u16,
 ) -> *mut line_sender_opts {
-    let builder = SenderBuilder::new(protocol.into(), host.as_str(), port);
+    let builder = SenderBuilder::new(protocol.into(), host.as_str(), port)
```
amunra, line 1438: I didn't get around that to why `SenderBuilder::new` returns a Result now, but this is a breaking change that will require a new full semantic version bump (if the API is indeed kept).

questdb-rs/src/ingress/mod.rs
```diff
@@ -601,20 +623,27 @@ impl SenderBuilder {
...
         let host = host.into();
         let port: Port = port.into();
         let port = port.0;
 
+        if host.is_empty() {
+            return Err(error::fmt!(ConfigError, "Empty host in address"));
+        }
+        if port.is_empty() {
+            return Err(error::fmt!(ConfigError, "Empty port in address"));
+        }
```
amunra, line 645: Pragmatically I think this can be reverted. If the host and port are invalid they can be validated later during the connection phase.

Keeping it requires a full version bump, which we probably do not want to do.

questdb-rs-ffi/Cargo.toml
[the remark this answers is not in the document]
amunra: This will be updated soon, I know, but for future reference you can use a `git` dependency here so you can run CI against your changes too.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories

questdb-rs-ffi/src/lib.rs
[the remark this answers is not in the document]
amunra: Coderabbitai: I don't believe this is an issue: The only requirement is that the swapped-in object can `Drop` cleanly without issues. The dummy object is never _actually_ used, unless there are bugs, of course. Feel free to double-check, but it should work as designed.

questdb-rs-ffi/src/lib.rs
amunra: See https://github.com/questdb/c-questdb-client/pull/138/changes#r2932706653

questdb-rs-ffi/src/lib.rs
```diff
@@ -1450,7 +1451,8 @@ pub unsafe extern "C" fn line_sender_opts_new_service(
     host: line_sender_utf8,
     port: line_sender_utf8,
 ) -> *mut line_sender_opts {
-    let builder = SenderBuilder::new(protocol.into(), host.as_str(), port.as_str());
+    let builder = SenderBuilder::new(protocol.into(), host.as_str(), port.as_str())
+        .expect("valid host and port");
```
amunra, line 1455: A `panic!` here is not acceptable, but I suspect you'll revert the Rust API change anyway.

questdb-rs/src/ingress/sender/http.rs
```diff
@@ -346,7 +380,7 @@ fn retry_http_send(
     let mut need_retry;
     loop {
         let jitter_ms = rng.random_range(-5i32..5);
-        let to_sleep_ms = retry_interval_ms + jitter_ms;
+        let to_sleep_ms = (retry_interval_ms + jitter_ms).max(0);
```
amunra, line 383: Why changed?
The jitter was designed to add a plus-or-minos adjustment.
Calling `.max(..)` disables the jitter about 50% of the time since the range starts at `-5`.
If this is intentional the correct change is probably to change the range passed into `random_range` above.
Did you encounter issues with the previous logic?

questdb-rs/src/ingress/sender/http.rs
```diff
@@ -357,6 +391,8 @@ fn retry_http_send(
             // see https://github.com/algesten/ureq/issues/94
             _ = last_rep.into_body().read_to_vec();
         }
+        // Rotate to the next address on retriable failure.
+        state.rotate();
```
amunra, line 395: This is a bit too simplistic, and probably wrong.
Even if it does work, it ends up wasting a lot of time during retries.

Issues:
* Always doing a roundrobin is _probably_ wrong here. The server could still be the primary. The round robin should happen if:
  * We know (or have reasonable suspicion) that the instance we contacted is likely offline (timeout, network error).
  * Or the instance responded with an error that confirms it's no longer primary.
  * In other words, it depends on the error.
* Round-robin is possibly needlessly timeconsuming.
  * What I mean here is that if a cluster has 4 addresses and the primary temporarily failed over to a replica and then comes back online and fails over to the original primary again (first on the list) then contacting replicas 3 and 4 isn't timely. Once successful I think it should start again from the start. Feel free to push back here. I may be wrong!

In short, we shouldn't create spurious network traffic to replicas we can avoid, especially since a buffer payload can be pretty darn big (even 100MiB).

questdb-rs/src/ingress/sender/http.rs
```diff
@@ -504,7 +540,7 @@ fn retry_http_get(
     let mut need_retry;
     loop {
         let jitter_ms = rng.random_range(-5i32..5);
-        let to_sleep_ms = retry_interval_ms + jitter_ms;
+        let to_sleep_ms = (retry_interval_ms + jitter_ms).max(0);
```
amunra, line 543: The fact that you changed this here too makes me wonder if the code can be refactored.

Same feedback applies.

questdb-rs/src/ingress/sender/http.rs
```diff
@@ -515,7 +551,9 @@ fn retry_http_get(
             // see https://github.com/algesten/ureq/issues/94
             _ = last_rep.into_body().read_to_vec();
         }
-        (need_retry, last_rep) = state.get_request(url, request_timeout);
+        // Rotate to the next address on retriable failure.
+        state.rotate();
+        (need_retry, last_rep) = state.get_settings(request_timeout);
```
amunra, line 556: Same feedback as before, plus refactor it so the logic isn't repeated.

questdb-rs/src/ingress/sender/mod.rs
```diff
@@ -192,22 +192,29 @@ impl Sender {
...
+                    request_timeout,
+                    retry_timeout,
                 ) {
                     Ok(res) => {
                         if res.status().is_client_error() || res.status().is_server_error() {
                             Err(parse_http_error(res.status().as_u16(), res))
                         } else {
-                            res.into_body();
+                            // Actively consume the body to return the connection to the pool.
+                            _ = res.into_body().read_to_vec();
```
amunra, line 210: My memory fades me, but IIRC reading to vec wasn't necessary. Please triple-check. If the user configured the wrong endpoint the response could be big.

amunra: Even if you do need to read, just read, no need to buffer into memory.

amunra: In general ask Claude to review if there's spurious unnecessary memory allocation in the flush or reconnect path: We've had bug reports from people who have been using the clients in trading engines who want to keep memory allocation to a minimum.

PR #139, opened 2026-04-24
test(ingress): meta-test that every column-writing Buffer method has an `_opt` twin

aaaronme:
## Summary

Follow-up to #134. Adds a meta-test in `tests/buffer_opt.rs` that scans `buffer.rs` and asserts every column-writing method has a matching `_opt` sibling.

Column writers are detected by the `N: TryInto<ColumnName<'a>>` generic bound rather than by a `column_*` name prefix — so `symbol` and any future unprefixed method (e.g. a hypothetical `uuid`) are also covered. If someone lands `column_uuid` or `uuid` without a `_opt` companion, the test fails with the missing names listed.

Two short regexes do the work; `[^{]*` matches across newlines so it spans multi-line `where` clauses without dotall. Adds `regex` to dev-dependencies (already pulled in transitively by other crates, so no new top-level dep at build time).

## Test plan

- [x] `cargo test --lib buffer_opt` — all 6 tests pass, including the new `test_every_column_writing_method_has_opt_variant`
- [x] Locally remove one `_opt` method from `buffer.rs` and confirm the test fails with that method name in the assertion message

## Summary by CodeRabbit

* **Tests**
  * Added internal validation test to ensure API consistency across methods.

* **Chores**
  * Updated development dependencies for testing infrastructure.

### DIFF
--- a/questdb-rs/Cargo.toml
+++ b/questdb-rs/Cargo.toml
@@ -58,6 +58,7 @@ chrono = "0.4.31"
 tempfile = "3"
 webpki-roots = "1.0.1"
 rstest = "0.26.1"
+regex = "1"
 
 [features]
 default = ["sync-sender", "tls-webpki-certs", "ring-crypto"]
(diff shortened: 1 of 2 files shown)


aaaronme:
@jerrinot wdyt? 
also considered CodeRabbits comment about excluding doc comments / examples

PR #141, opened 2026-05-11
feat(ingress): QWiP protocol implementation

jerrinot:
QWiP over UDP and Websocket transports.
The Websocket transport includes a durable Store and Forward mechanism. 

## FFI panic-safety hardening

`line_sender_buffer_reserve` and `line_sender_buffer_clone` previously propagated Rust panics across the `extern "C"` boundary (capacity overflow in `Vec::reserve`, allocation failure in `Buffer::clone`), aborting the process with no recovery path for C callers.

Both now NULL-guard their buffer argument and wrap their body in `catch_unwind`, surfacing failures via the standard `err_out` channel:

```c
// before
void               line_sender_buffer_reserve(line_sender_buffer*, size_t);
line_sender_buffer* line_sender_buffer_clone (const line_sender_buffer*);

// after
bool                line_sender_buffer_reserve(line_sender_buffer*, size_t, line_sender_error**);
line_sender_buffer* line_sender_buffer_clone (const line_sender_buffer*, line_sender_error**);
```

### Breaking changes for C++ consumers

The C++ wrapper translates the new error channel into thrown `line_sender_error` exceptions, matching the rest of the C++ API (`bookmark()`, etc.). As a side effect, the following members **lose their `noexcept` specification**:

- `line_sender_buffer::line_sender_buffer(const line_sender_buffer&)` (copy constructor)
- `line_sender_buffer::operator=(const line_sender_buffer&)` (copy assignment)
- `line_sender_buffer::qwp_udp(...)` (static factory)

Downstream code that relies on `std::is_nothrow_copy_constructible_v<line_sender_buffer>` (e.g. for the strong-exception-guarantee fast path in `std::vector<line_sender_buffer>` resize) will observe a different value. Move construction / move assignment remain `noexcept`.

C consumers must now check the `bool` return of `line_sender_buffer_reserve` (the 12 in-tree `examples/*.c` files were updated accordingly) and pass `&err` to `line_sender_buffer_clone`.

## Summary by CodeRabbit

* **New Features**
  * Added QWP/UDP and QWP/WebSocket transports, unified synchronous Sender APIs, buffer bookmark/rewind support, FSN progress/ack controls, structured server-rejection diagnostics, and new language examples & benchmark.

* **Documentation**
  * Added QWP/WebSocket architecture, API/FFI specs, error-handling/progress design docs, security guidance, examples manifest, and updated README/ingress docs.

* **Chores**
  * Pre-commit hook (format/lint), expanded .gitignore, build/example targets, CI pipeline tweaks, and added tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/questdb/c-questdb-client/pull/141)


bluestreak01:
@jerrinot — full review below. Level 2 pass (8 parallel review agents + batched verification); level 3 is available for a deeper sweep if you want it. **Verdict: request changes**, primarily on missing E2E coverage and a few correctness/resource issues.

---

# Review summary

**Scope:** 61,471 added lines across 102 files. QWP/UDP (best-effort datagrams) and QWP/WebSocket (with durable Store-and-Forward) transports. Public Rust API, C ABI (`line_sender.h` + ~25 new `line_sender_qwpws_*` functions), C++ wrapper, FFI bindings.

This is a substantial, broadly well-engineered addition — careful overflow-checked arithmetic, disciplined `unsafe`, `qwpws_catch_bool` envelopes on most new FFI entry points, sound lock-free `published_upper`/`completed_upper` discipline with verified Acquire/Release pairing.

But it ships with **zero real-server `system_test/` coverage for QWP/WebSocket, zero TLS coverage, zero auth coverage, and no QWP/WS example program** — all four are this project's hard prerequisites for shipping a wire-protocol transport. Add an orphan-drainer thread leak past sender Drop (C5), a `close_drain` budget that doesn't bound real close cost (M7), and the team's own self-review ledger documenting that high-priority protocol-parity gaps (`SELF_REVIEW_TODOS.md` R1/R3/R5) are still open — and this isn't ready.

**Findings stats:** 35 verified, 8 dropped as false positives. **5 P0** (all out-of-diff or shipping-readiness), **24 P1**, **14 P2**.

---

## Critical (blocks merge)

### C1. Zero `system_test/` coverage for QWP/WebSocket — out-of-diff
**`system_test/test.py`, `system_test/fixture.py`, `system_test/questdb_line_sender.py`**
`grep` for `qwpws|qwp_ws` over `system_test/` returns nothing. The Python wrapper enum stops at `QWPUDP`. `fixture.py` doesn't start QuestDB with `qwp.ws.*` enabled. Per project standard ("Real-server E2E coverage is MANDATORY for protocol/auth/TLS/lifecycle/failover/public-ingest changes"), a wire-protocol addition can't ship without `system_test/` coverage. **Fix:** add `TestQwpWsSender` and `TestQwpWssSender` covering round-trip, schema evolution, reconnect+replay, durable ACK ordering, server-side reject (drop-and-continue + halt), terminal close.

### C2. Zero `qwpwss` (TLS) test coverage
`grep -rn "QwpWss\|qwpwss"` over `questdb-rs/src/tests/`, `system_test/`, `cpp_test/` returns only the config-parse acceptance test (`ingress/tests.rs:174`). No connection, handshake, certificate, or `tls_proxy` test. TLS is a release blocker for any production transport.

### C3. Zero authentication test for QWP/WS
No test sends/validates the `Authorization:` header on the WebSocket upgrade. Mock servers in `tests/qwp_ws.rs` don't assert it. Customers will silently authenticate or silently fail.

### C4. No QWP/WS example program — out-of-diff (`examples/`)
`ls examples/` shows 4 `qwpudp` examples (C + C++, batch + simple) and 0 `qwpws` examples. The QWP/WS API is significantly more complex than UDP (`flush_and_get_fsn`, `await_acked_fsn`, `poll_qwp_ws_error`, `close_drain`, `drive_once`); customers can't adopt it from the header alone.

### C5. Detached orphan-drainer threads leak flock + FD + TCP — `qwp_ws_orphan.rs:187-200`
`OrphanDrainerPool::close_with_timeouts` waits `ORPHAN_POOL_GRACEFUL_DRAIN + ORPHAN_POOL_STOP_GRACE` (~3 s), then `self.threads.clear()` drops `JoinHandle`s without joining. Each detached worker still owns a `BlockingQwpWsTransport` (TCP + optional TLS) and an `SfaSlotQueue` holding an `flock`'d lock FD and one or more `MmapMut` mappings. On a slow/hostile network the thread can live indefinitely past sender Drop, blocking re-adoption of the same orphan slot. Your handover doc `QWP_WEBSOCKET_ORPHAN_DRAINING_TODO.md:37` flags this is "unresolved". **Fix:** before `clear()`, `TcpStream::shutdown(Both)` via a stored `try_clone` so blocked syscalls unblock; then join.

---

## Moderate (P1)

### M1. `line_sender_flush` / `_keep` / `_keep_with_flags` not wrapped in `catch_unwind` — `questdb-rs-ffi/src/lib.rs:2819, 2847, 2883`
Pre-existing structure, but this PR extends these entry points to QWP/WS, which involves a background runner thread with poisoned-mutex possibilities and panic sites that didn't exist before. QWP/WS-specific entries correctly use `qwpws_catch_bool`; the bare flush family doesn't. A panic crossing FFI is UB (process abort in C, worse in Cython). Wrap all three bodies.

### M2. `await_acked_fsn` busy-polls instead of using `BackpressureNotifier` — `questdb-rs/src/ingress/sender.rs:561-591`, `qwp_ws.rs:425-451`
In background mode the loop sleeps 50 µs and re-reads `acked_fsn`, taking the publication mutex ~20,000×/sec. The runner already calls `BackpressureNotifier::notify_all` on every ACK/REJECT/sent-progress/storage change (`qwp_ws.rs:537, 572, 656, 686, 731, 784, 801, 816`). `close_drain` correctly uses `wait_for_change` (`qwp_ws.rs:373-391`); `await_acked_fsn` should do the same.

### M3. `await_acked_fsn(fsn > published_fsn)` silently hangs to timeout — `sender.rs:572-590`
No upper-bound check against `published_fsn()`. Customer code mixing up FSN values across sender instances burns wall time. Same loop: `Instant::now().checked_add(timeout)` returns `None` for `Duration::MAX`, and `qwp_ws_deadline_expired(None)` returns `false` → infinite wait. Document the latter explicitly or reject the former.

### M4. Server ACK `wire_seq > last_sent_wire_seq` silently capped — `qwp_ws_driver.rs:1785-1797`
`ack_fsn_for_wire_seq` and `reject_fsn_for_wire_seq` do `wire_seq.min(last_sent_wire_seq)` and then advance the completion watermark to the cap-adjusted FSN. A misbehaving or fuzzed server can quietly make the client believe data was durably ACKed that it never sent. Error as `ProtocolViolation`, don't cap. (Also flagged in your own `SELF_REVIEW_TODOS.md` R3.)

### M5. `SyncQwpWsRunner::acked_fsn` locks publication mutex to read an atomic — `qwp_ws.rs:337-341`, `qwp_ws_sfa_queue.rs:1014-1019`
`completed_fsn()` only reads `completed_upper: AtomicU64`. The mutex is unnecessary; contended hot path under the same lock as the runner's progress steps.

### M6. SFA segment + hot-spare files persist on disk on non-drained Drop — `qwp_ws_sfa_queue.rs:805-820`
`SfaEngine::close()` calls `record_all_sfa_cleanup` only when `fully_drained` is true. A sender dropped mid-flight (panic, terminal error, user abandonment) leaves `.sfa` segments + the preallocated hot-spare under `sf_dir` forever — no boot-time TTL, no operator purge. Per-`sender_id` accumulation across crashes. Document + add an operator-facing purge knob.

### M7. `close_drain` timeout doesn't terminalise the runner — `qwp_ws.rs:2028-2036`, `qwp_ws.rs:901-908`
On timeout, `close_drain` returns `SocketError` but doesn't call `lifecycle.terminalize()` or shut down the underlying `TcpStream`. The runner keeps running in `Closing`; subsequent `Drop` joins the thread, which can stall up to `request_timeout` if a `write_all` is blocked. The caller's close-drain budget is therefore not actually the close budget. **Fix:** on timeout, terminalize lifecycle + `TcpStream::shutdown(Both)`, then join. (Also `SELF_REVIEW_TODOS.md` R5.)

### M8. `crc32c` and `memmap2` are non-optional `questdb-rs` dependencies — `questdb-rs/Cargo.toml:44-45`
Both are only used under `_sender-qwp-ws`. ILP-only consumers now compile them. Mark `optional = true` and gate behind the appropriate feature.

### M9. 5 QWP/WS modules use `#![allow(dead_code)]` at file scope
`qwp_ws_driver.rs:25`, `qwp_ws_queue.rs:25`, `qwp_ws_sfa_queue.rs:25`, `qwp_ws_sfa_segment.rs:25`, `qwp_ws_sfa_slot.rs:25`. File-scoped allows hide unused-code accumulation across this 20K-line addition. Replace with targeted attributes or delete the dead items.

### M10. New C header types are undocumented — `include/questdb/ingress/line_sender.h:881-924`
Seven types/enums (`line_sender_qwpws_progress`, `line_sender_qwpws_fsn`, `line_sender_qwpws_error_category`, `line_sender_qwpws_error_policy`, `line_sender_qwpws_error`, `line_sender_qwpws_error_view`) have zero `/** ... */` doc blocks. Pre-existing enums (`line_sender_error_code`, `line_sender_protocol`) document every variant. Also: the header doesn't say `line_sender_qwpws_error_view.message` is not null-terminated — a Cython caller doing `PyUnicode_FromString(view.message)` crashes.

### M11. New variant naming inconsistent with existing C header
Pre-existing variants are `snake_case` (`line_sender_protocol_tcp`, `line_sender_error_invalid_decimal`). New QWP/WS variants are `SCREAMING_SNAKE` (`LINE_SENDER_QWPWS_PROGRESS_BACKGROUND`, `LINE_SENDER_QWPWS_ERROR_HALT`). Worse: `LINE_SENDER_QWPWS_ERROR_*` is the prefix for *both* the `*_error_category` enum AND the `*_error_policy` enum, so `_HALT` (policy) reads like an error category. Rename to `LINE_SENDER_QWPWS_CATEGORY_*` and `LINE_SENDER_QWPWS_POLICY_*` (or align with snake_case).

### M12. No CHANGELOG entry, no version bump
`questdb-rs/Cargo.toml:3` and `questdb-rs-ffi/Cargo.toml:3` both still `version = "6.1.0"`. New transports, new public Rust API, ~25 new C functions, 7 new C types — at minimum a minor bump (6.2.0). No `CHANGELOG.md` exists or is added.

### M13. README does not mention QWP/WebSocket
`README.md` + `questdb-rs/README.md` advertise only ILP/HTTP, ILP/TCP, QWP/UDP. QWP/WS is undocumented for first-time readers despite ~25 public C functions. Same gap in `doc/C.md` and `doc/CPP.md`.

### M14. Ships with documented high-priority open TODOs — `doc/QWP_WEBSOCKET_SELF_REVIEW_TODOS.md`
The PR itself ships a TODO ledger listing **R1 (high)** "Make Rust QWP/WebSocket response parsing match Java/server framing — parser accepts truncated/trailing-garbage frames Java rejects", **R3 (medium)** "Align ACK/NACK beyond highest sent with Java clamping" (cf. M4), and **R5 (high)** "Make runner shutdown and close/drain behavior explicit and Java-compatible" (cf. M7 and C5). These are protocol-correctness gaps you flagged yourselves.

### M15. C++ `flush_and_keep` / `_with_flags` silently substitute an empty buffer when `buffer._impl == nullptr` — `include/questdb/ingress/line_sender.hpp:1422-1436, 1461-1482, 1519`
The C API treats NULL buffer as UB per the header preamble. The C++ wrapper instead calls `this->new_buffer(0)` and flushes an empty buffer — silently issuing an HTTP request (or a no-op UDP send) for what should be an error. Two-sided behaviour diverges from the C contract; Python (which goes through the C path) won't see this fallback.

### M16. C++ `new_buffer` doesn't check NULL from `line_sender_buffer_new_for_sender` — `line_sender.hpp:1351-1368`
Followed immediately by `line_sender_buffer_reserve(raw_buffer, init_buf_size)`. If the C call ever returns NULL (today only on allocation failure, but the header doesn't promise non-NULL), the C++ side dereferences NULL.

### M17. Per-row column lookup allocates when columns are out of order — `questdb-rs/src/ingress/buffer/qwp.rs:3050-3062, 3730-3739`
The cursor fast-path correctly skips this. The fallback path (`column_lookup_key`) does `std::str::from_utf8(name)?.to_lowercase()` (heap-allocates a `String`) and then probes a `HashMap<String, usize>` with default SipHash on every per-row column write whose order doesn't match the order columns were first created. Optional or skipped-column workloads pay this every row. **Fix:** key the map on `Vec<u8>` lowercased bytes with a fast hasher; probe with a stack scratch for short names; reuse the already-stored `lower_ascii_name` cache.

### M18. Per-row `Vec<u8>` allocation for QWP/WS array columns — `qwp.rs:2543-2563, 2824-2856`
`encode_array_value` returns an owned `Vec<u8>` per row, then `append_double_array` copies it into the column data Vec. Two memcpys + one alloc per array column per row. The QWP/UDP path encodes directly into the column data Vec (`append_value_f64_array`, `qwp.rs:1132`). Mirror that for QWP/WS.

### M19. False sharing on `published_upper` / `completed_upper` — `qwp_ws_sfa_queue.rs:247-256`
Producer hot-writes `published_upper`; consumer (runner) hot-writes `completed_upper`; they sit 8 bytes apart on the same 64-byte cache line. Producer reads `completed_upper` in `validate_submit` every submit, bouncing the line off the runner. Same pattern on `SfaSharedSegment` at `qwp_ws_sfa_queue.rs:1111-1112`. `grep` for `CachePadded|repr(align(64))` over the whole tree returns zero. **Fix:** `crossbeam_utils::CachePadded` on every cross-thread atomic written by one party and read/written by another.

### M20. WebSocket frame masking copies + scalar XORs payload — `qwp_ws_codec.rs:154-159`
Per-frame `out.extend_from_slice(payload)` then `for b in iter_mut() { *b ^= mask[i & 3]; }`. (a) `send_buf` initial capacity is 16 KiB (`qwp_ws.rs:1899`) which is below realistic batch frame sizes — triggers regrowth allocs; (b) the masking loop is per-byte. Bump initial capacity, mask in `u64` chunks with a broadcast mask.

### M21. No `questdb-rs/benches/` directory; no Criterion microbench for any QWP hot path
The single `questdb-rs/examples/qwp_ws_unified_sfa_bench.rs` is a one-off main requiring a live server. Hot paths added by this PR with no microbench: QWP/UDP encoding, QWP/WS columnar encoding, QWP/WS row build, `QwpWsReplayEncoder::encode_with_max_size`, `SfaFrameQueue::try_submit` lock-free path, driver send loop. **Per project standard, missing benches for hot-path code blocks merge.**

### M22. `unsafe { ... }` blocks missing `// SAFETY:` comments
- `questdb-rs/src/ingress/buffer/ilp.rs:60-73` — three `unsafe` writes via `set_len` + `get_unchecked_mut` depend on `check_escape_fn` being deterministic across the count pass and the write pass; no documentation. A future caller passing a non-pure closure causes OOB writes.
- `questdb-rs/src/ingress/sender/qwp_ws_sfa_slot.rs:244` — `unsafe { libc::flock(...) }` with no annotation. Trivial fix.

### M23. Bugfix regression test dropped — disjoint-payload Miri test for `qwp_ws_queue.rs`
Commit `516f805` "Fix QWP WS queue payload aliasing" added a Miri-oriented regression for publishing a disjoint payload while a previous view is live. Later refactors (`85303d9`, `4edac7e`) appear to have dropped it: `grep -n "disjoint\|alias" questdb-rs/src/ingress/sender/qwp_ws_queue.rs` returns nothing. Re-introduce it.

### M24. Multi-datagram QWP/UDP partial flush re-sends rows on retry — `sender.rs:494-498`, `qwp.rs:1738-1824`
`flush_to_socket` returns on first `send()` error; `Sender::flush` only clears the buffer on Ok. If `send()` of datagram k succeeds and k+1 fails, the customer retries the whole batch and datagrams 1..k go again. At-least-once semantics for UDP is reasonable, but the docs don't say so. Document explicitly or track a send cursor.

---

## Minor (P2)

- **`SymbolGlobalDict` (`qwp.rs:3776-3845`)** grows monotonically per distinct symbol value; `reset()` exists but is `#[allow(dead_code)]`. High-cardinality symbol columns OOM eventually. Bounded by user behaviour, but expose a reset hook on flush/reconnect.
- **`line_sender_qwpws_error_get_view` / `line_sender_error_qwpws_get_view` not in `catch_unwind`** (`lib.rs:2687-2723`). Both bodies are provably panic-free field copies; safety risk is hypothetical. Wrap for consistency with the rest of the QWP/WS FFI surface.
- **`line_sender_opts_max_datagram_size` and `_multicast_ttl`** (`lib.rs:1705-1726`) deref NULL `opts` unconditionally; the new `_qwpws_progress` at 1729 NULL-checks. Pick one convention.
- **Driver loops accept negative `seq_txn` from server** — `qwp_ws_driver.rs:2441-2453`, `qwp_ws_codec.rs:471`. Defense-in-depth: validate `seq_txn >= 0`.
- **`qwp.rs:2250`** uses unchecked `u32` `entry.offset + entry.len`; sister function uses `checked_add`. Internal-only data, low risk.
- **`qwp_ws_sfa_segment.rs:249`** `size_bytes as usize` truncates on 32-bit (error-path only).
- **`tests/qwp_ws.rs:2507-2527`** `qwp_ws_from_conf_parses_java_reconnect_keys` doesn't read back any parsed field. Tighten.
- **25 `thread::sleep` calls in `tests/qwp_ws.rs`**, mostly 50–100 ms post-write delays to let ACKs land. Flake risk on loaded CI; replace with explicit sync.
- **Missing `#[inline]` on per-row `QwpWsColumnBuffer::append_*`** (`qwp.rs:3168-3289`). Public per-row entry points are `#[inline(always)]`; the inner appenders aren't.
- **`.codex`** 0-byte file committed at repo root (IDE artifact). Remove.
- **Typo** "maxinum" → "maximum" in `sender.rs:728` rustdoc.
- **30 design-journal `.md` files** added under `doc/QWP_WEBSOCKET_*.md`. Developer notes, not user docs. Move to `doc/internal/` or a wiki.
- **Undocumented `ws`/`wss` config-string aliases** in `ingress.rs:363, 365`. Either expose in C header or remove.
- **Drop blocking on in-flight `write_frame`** — `qwp_ws.rs:901-908`. No `TcpStream::shutdown` on stop; `Drop` may stall up to `request_timeout`. Companion to C5/M7.
- **Rust API names** like `Sender::flush_and_get_fsn`, `await_acked_fsn`, `drive_once`, `close_drain` aren't `qwp_ws_`-prefixed but return `InvalidApiCall` at runtime for non-WS senders. Other methods correctly use `qwp_ws_*` prefix (e.g. `poll_qwp_ws_error`). Inconsistent.

---

## Downgraded (false positives / overstated)

- **"Silent durability loss with `sf_durability=memory` + `sf_dir`" (initially flagged P0/P1).** The code at `qwp_ws.rs:910-944` rejects every non-`Memory` durability mode, so memory-mode is the only mode today and it correctly advertises memory semantics. On-disk segments are a recovery aid, not a durability guarantee, so no `msync`/`fsync` is required. **Downgraded to P2 doc gap** in `QWP_WEBSOCKET_DURABLE_SF_STORAGE.md`.
- **"`complete_through` doesn't guard backward-moving watermark".** `qwp_ws_sfa_queue.rs:843-857` does guard: `target_upper <= completed` short-circuits, `> published` errors. Stale-ACK is a no-op.
- **"`drive_once` in background mode → InvalidApiCall not tested".** Heavily tested in `tests/qwp_ws.rs:1144, 1145, 1237, 1238, …`.
- **"`line_sender_qwpws_error_get_view` / `line_sender_error_qwpws_get_view` are P0 panic-catch bugs".** Verified panic-free; reclassified P2 (consistency).
- **"`published_frame_count(&self)` is a data race".** Currently safe — the only caller goes through `published_frame_count_after_offset` which Acquires `published_offset` first. Hypothetical-future-caller hazard; P2 cleanup.
- **"All real-server QWP/WS probes are `#[ignore]`-gated" framed as P0.** Probes exist; CI just doesn't wire them. Reclassified P1 (CI activation needed).
- **"Per-row `to_lowercase()` is P0".** Cursor fast-path skips it. P1 for reordered-column workloads (see M17).
- **Pre-existing NULL-deref convention on `unwrap_sender*` / `unwrap_buffer*`.** Not a PR regression; carried over from before. Header preamble documents NULL as UB. Skipped.

---

## Summary checklist

- **Real-server `system_test/` coverage for QWP/WebSocket:** ❌ absent — blocks merge (C1).
- **TLS / `qwpwss` test coverage:** ❌ absent — blocks merge (C2).
- **Auth test coverage for QWP/WS:** ❌ absent — blocks merge (C3).
- **QWP/WS example program for customers:** ❌ absent — blocks merge (C4).
- **Hot-path Criterion benchmarks:** ❌ absent — blocks merge (M21).
- **CHANGELOG + version bump:** ❌ absent (M12).
- **FFI surface vs C header:** ✅ matches (no symbol drift). Naming and docs need cleanup (M10, M11).
- **N:M threading model:** ✅ preserved.
- **Streaming contract:** N/A (this PR is ingress only).
- **No P0 Rust safety / soundness bugs in new code.** Care taken with overflow-checked arithmetic and `unsafe` is genuinely good.

Out-of-diff findings are concentrated in `system_test/`, `examples/`, `README.md`, and the C++ wrapper — exactly where Cython/Python and end-user adoption begins.

Happy to re-review once `system_test/` + examples + TLS/auth coverage land, or to do a level-3 pass if you'd prefer a deeper sweep before that.

bluestreak01:
@jerrinot — code review below (level 2, automated multi-agent pass against `main`).

**Level 2 review** — level 3 exists for the full mission-critical pass; recommend running it before merge given the size (57,991 additions / 4,418 deletions / 80 files, including a 9,534-line buffer, 6,874-line driver, and a new lock-free Store-and-Forward queue).

## PR description / housekeeping

- The body is one human-authored sentence plus the CodeRabbit auto-summary. For a PR of this size, the description should explicitly call out: new public C functions (18) and new error/protocol enum variants; modified C ABI structs (`line_sender_error`, `line_sender_buffer`); the new on-disk Store-and-Forward area; threading and `Drop` semantics for the QWP/WS sender; cleanup behavior on crash (orphan lock files). No issue references; no migration notes for downstream bindings.
- Naming inconsistency between **PR title** ("QWiP") and **codebase** ("QWP"/"qwp_ws"/"qwpws"). The implementation itself is inconsistent: Rust uses `qwp_ws_*`, C uses `qwpws_*`, conf strings mix `_ms`/`_millis`. Pick one and stick to it before merging.

## Critical (must fix before merge)

- ~~**C1 — ABI break: `line_sender_error_server_rejection` inserted in the middle of `line_sender_error_code`.**~~ ✅ Fixed in 482120a (ServerRejection moved to end of ErrorCode).
  `include/questdb/ingress/line_sender.h:89` and `questdb-rs/src/error.rs:72-74`. Verified: pre-PR ordering had `server_flush_error, config_error, array_error, protocol_version_error, invalid_decimal`; the PR splices `server_rejection` between `server_flush_error` and `config_error`. Discriminants for `config_error`, `array_error`, `protocol_version_error`, and `invalid_decimal` all shift +1. Every downstream language binding caching numeric values (questdb-py via ctypes/cffi, Go cgo, Java FFM) silently misroutes errors after this. **Fix:** Append `ServerRejection` after `InvalidDecimal`; even better, pin discriminants with explicit `= N` in the C enum.

- ~~**C2 — `line_sender_buffer_reserve` and `line_sender_buffer_clone` propagate panics across the FFI boundary.**~~ ✅ Fixed in 482120a (NULL guards + `catch_unwind` + `err_out`).
  `questdb-rs-ffi/src/lib.rs:909-918` (reserve) and `:887-899` (clone). Both call into `Vec::reserve` / `Buffer::clone()` with no `catch_unwind`. `Vec::reserve(SIZE_MAX/2)` panics on capacity overflow; clone has 5 `to_vec` allocations. Panic unwinding across `extern "C"` is UB (modern rustc aborts). `line_sender_buffer_clone(NULL)` also has no NULL guard at the call to `unwrap_buffer`. **Fix:** Add NULL guard and wrap body in `catch_unwind` (or move to `try_reserve` and return `false`).

- ~~**C3 — New bookmark FFI exports lack NULL checks on `buffer`.**~~ ❌ Declined in bf9243d. Same pattern applies to every `line_sender_buffer_*` function (and the sender-handle family); the rule is a library-wide C convention. Documented explicitly at `include/questdb/ingress/line_sender.h:41-60`. No panic-safety risk in `clear_bookmark` (no allocation), so `catch_unwind` is not warranted; only NULL-handle behavior is at issue, and that's now stated as caller responsibility.
  `questdb-rs-ffi/src/lib.rs:935-953` (`line_sender_buffer_bookmark`), `:957-968` (`rewind_to_bookmark`), `:971-980` (`clear_bookmark`). `unwrap_buffer_mut` at lib.rs:882-884 dereferences without a NULL check. `clear_bookmark` is `void`-returning, so a buggy caller cannot recover. **Fix:** Add `if buffer.is_null() { … }` mirroring `line_sender_buffer_free`.

- ~~**C4 — `upd_opts!` macro is not wrapped in `catch_unwind`.**~~ ✅ Fixed in 1e0a860 (macro body now wraps in `catch_unwind`, covering all 20+ opts setters at once; redundant `qwpws_catch_bool` + explicit NULL check removed from `qwpws_progress` so every setter has the same one-line body and matches the documented NULL-handle convention).
  `questdb-rs-ffi/src/lib.rs:171-187`. The macro does `builder_ref.clone().$func(...)` and is used by every new QWP setter (`line_sender_opts_max_datagram_size`, `multicast_ttl`, `qwpws_progress` at lib.rs:1700-1738). `SenderBuilder::clone` allocates; setters can `format!` for errors. Any panic unwinds across the FFI boundary → UB. **Fix:** Wrap the macro body in `catch_unwind` or migrate the new setters to the `qwpws_catch_bool` family.

- ~~**C5 — User-supplied error-handler `cb` invoked without `catch_unwind`; `line_sender_flush` family lacks `catch_unwind` too.**~~ ✅ Fixed in 9ec413a. Two narrow helpers: `ffi_catch_bool` (renamed/generalized from `qwpws_catch_bool`) now wraps every bool-returning extern "C" function uniformly, including the three legacy flush entry points. `ffi_invoke_c_callback` wraps the user-callback trampoline (panic logged via `io::stderr` and swallowed — no `err_out` channel for void callbacks). Every outbound and inbound FFI boundary now has the same shape.
  Trampoline at `questdb-rs-ffi/src/lib.rs:1768-1807` stores `cb(user_data, &view)`. Invoked via `QwpWsErrorHandler::handle` from `drain_qwp_ws_error_notifications` (sender.rs:220-235), reachable from `line_sender_flush` (lib.rs:2910-2926) and `flush_and_keep[_with_flags]` (lib.rs:2938, 2974), none of which wrap in `catch_unwind`. A C++ callback throwing or a Python ctypes callback raising unwinds across two `extern "C"` boundaries. **Fix:** Wrap the trampoline `cb` call in `catch_unwind(AssertUnwindSafe(...))`; wrap all flush entry points.

- **C6 — `QwpWsColumnarBuffer.snapshots` grows unboundedly.**
  `questdb-rs/src/ingress/buffer/qwp.rs:2330-2362`. `capture_snapshot` pushes a deep-cloned snapshot; `restore_snapshot` only `.get(idx).cloned()`; `clear()` clears table rows and the bookmark but never truncates `self.snapshots`. Idiomatic per-flush "set_marker, build, flush, repeat" usage leaks the prior snapshot's heap each cycle. The QWP/UDP `QwpBuffer` correctly uses inline watermark-based markers. **Fix:** Truncate `self.snapshots` in `restore_snapshot` and `clear`; or rewrite to use per-table watermarks like `QwpMarker`.

- **C7 — No `connect_timeout`, plus orphan-drainer thread detach can leak FDs, TCP/TLS sessions, slot locks, and mmaps.**
  `questdb-rs/src/ingress/sender/qwp_ws.rs:2144-2169`. `connect_tcp_to_any_addr(host, port, addrs, request_timeout)` accepts `request_timeout` but calls `TcpStream::connect(addr)` (line 2152) **without** it; the timeout is applied only as read/write timeout after the connection succeeds. On Linux that's ~75s SYN retry. `Sender::Drop` joins the runner thread, which can't return until the connect returns. `OrphanDrainerPool::close_with_timeouts` (qwp_ws_orphan.rs:191-199) waits ~3s then `self.threads.clear()` to detach — detached threads continue holding `SlotLock`/`.lock` File + TCP/TLS sessions + ack-watermark mmaps. With network partitions + repeated `Sender::from_conf`, this leaks unboundedly. **Fix:** Use `TcpStream::connect_timeout(addr, request_timeout)`; before joining, `shutdown(SHUT_RDWR)` the socket so blocked I/O unblocks; bound orphan retry budget.

## Moderate (verified, worth addressing)

- **M1 — i32 cumulative-offset overflow corrupts wire for >2 GiB string columns.** `buffer/qwp.rs:3600-3607` and `:5471-5482`. Both paths accumulate per-cell length into `let mut cumulative: i32 = 0` without `checked_add`. Encoding runs before any `max_buf_size` check (qwp_ws_publisher.rs:62-78), so a user with `max_buf_size > 2 GiB` plus large text columns gets silent wire corruption. **Fix:** Track in `i64`/`u64`, validate against `i32::MAX` before writing the `i32` offset.
- **M2 — Asymmetric panic protection across `extern "C"`.** `line_sender_flush`/`flush_and_keep[_with_flags]`, `line_sender_get_protocol`, `line_sender_buffer_new_for_sender` all dereference `sender` unchecked and have no `catch_unwind`. Same underlying panic produces UB through one entry point and a clean `bool false` through another. **Fix:** Add NULL guards and `catch_unwind` everywhere, or document the abort-on-panic rule and remove `qwpws_catch_bool`.
- **M3 — `line_sender_qwpws_error_get_view(NULL)` returns a fabricated UNKNOWN/HALT view.** `questdb-rs-ffi/src/lib.rs:2778-2797`. Indistinguishable from a real UNKNOWN/HALT diagnostic. **Fix:** Return via out-parameter + bool, or document the sentinel discriminants distinctly.
- **M4 — `parse_qwp_ws_endpoints` cannot parse bracketed IPv6 literals.** `questdb-rs/src/ingress.rs:455-456`. `entry.split_once(':')` mangles `[::1]:9000`. Any IPv6-only host is broken with a misleading "invalid port" error. **Fix:** Special-case `[...]:port`.
- **M5 — Orphan-drainer threads detached without socket shutdown.** See C7 — same root cause; the leak is unbounded.
- **M6 — Asymmetric `catch_unwind` between `line_sender_qwpws_error_free` and `line_sender_error_free`.** lib.rs:2816-2823 vs :442-449. Same operation, different protection.
- **M7 — Aliasing UB: FFI synthesises `&Sender` and `&mut Sender` over the same raw pointer with no documented thread-safety contract.** `qwpws_line_sender_ref` (lib.rs:2284-2300) and `qwpws_line_sender_mut` (:2302-2318). `published_fsn`, `acked_fsn`, `protocol` take `&self`; `flush_and_get_fsn` takes `&mut self`. C header documents no concurrency contract. **Fix:** Document single-thread or external-mutex requirement in the C header.
- **M8 — `set_err_out_from_sender_error` heap-allocates on every non-QWP/WS sender failure.** lib.rs:519-533 calls `sender.qwp_ws_terminal_error()` which `Err`s with a heap message dropped by `.ok()`. **Fix:** Gate on `sender.protocol()`.
- **M9 — WS payload masking is per-byte.** `qwp_ws_codec.rs:155-160`. 4–8× slower than chunked u64 XOR; trivially fixable.
- **M10 — `RowGroupPlanner::find_column` is O(C) inside per-row loop → O(N·C²) per buffer.** `buffer/qwp.rs:5133` for QWP/UDP. QWP/WS has a column cursor; QWP/UDP does not. **Fix:** Add `HashMap<&[u8], u16>`.
- **M11 — `QwpWsColumnarBuffer::column_arr` double-allocates per array row.** `buffer/qwp.rs:2543-2562, :2856`. **Fix:** Mirror QWP/UDP — write directly into the column arena.
- **M12 — `set_marker` deep-clones every table and column.** `buffer/qwp.rs:2339-2348`. `set_marker` per-row is O(buffer-size). **Fix:** Per-table watermarks like `QwpMarker`.
- **M13 — Three `.expect("durable ACK mode")` in driver hot paths.** `qwp_ws_driver.rs:828, 843, 1031`. Panic payload is silently dropped by `let _ = thread.join()`. Refactor hazard. **Fix:** Replace with `internal_error!` returns.
- **M14 — Direct callers of `SfaSharedSegment::published_frame_count()` bypass the `published_offset` Acquire barrier.** `qwp_ws_sfa_queue.rs:1029, 1847`. On aarch64/POWER they can observe stale counts; benign on x86 TSO. **Fix:** Route all readers through `published_frame_count_after_offset`, or make the producer use `Release` on `frame_count`.
- **M15 — `complete_through_fsn` is non-atomic read-modify-write of `completed_upper`.** `qwp_ws_sfa_queue.rs:905-919`. Safe today (single caller). **Fix:** `fetch_max(target_upper, AcqRel)`.
- **M16 — `wire_seq` clamp silently hides server protocol violations.** `qwp_ws_driver.rs:1988-2014`. `QueueError::ProtocolAckBeyondSent` is declared but never produced; an ACK targeting an unsent frame is re-mapped to the highest in-flight. **Fix:** Surface the protocol-violation error.
- **M17 — `c_default_qwp_ws_error_handler` uses `eprintln!` which panics on broken stderr.** lib.rs:2221-2245. Installed by default for every opts construction. **Fix:** `let _ = writeln!(io::stderr(), ...)` or `log::error!`.
- **M18 — TLS sessions dropped without `close_notify`.** Peer sees unclean shutdown.
- **M19 — `SyncProtocolHandler::Drop` only `shutdown(Write)` for SyncTcp.** tcp.rs:144-158 — the existing rationale (Windows queued bytes) was not extended to the new QWP/WS variant.
- **M20 — Hot-spare segment file leaked on `allocated_segment_bytes` overflow.** `qwp_ws_sfa_queue.rs:1037`. Astronomically unlikely but trivially preventable with `SfaStorageCleanup::new(...)` before the `?`.
- **M21 — `line_sender_qwpws_flush_and_get_fsn` runs the stale-error-notification path before validating the buffer flavor.** ILP buffer + QWP/WS sender → user sees a confusing server-error notification, then the API misuse error. **Fix:** Validate first.
- **M22 — `line_sender_buffer_peek` returns NULL for empty QWP and non-NULL empty for ILP — undocumented.** lib.rs:1093-1108. The new `empty_peek_buf_is_null` flag creates protocol-dependent behavior. **Fix:** Document explicitly or unify.
- **M23 — All twelve new `line_sender_qwpws_*` C functions lack Doxygen `@param`/`@return`.** line_sender.h:1393-1521, plus types at 884-927. Inconsistent with the rest of the header.
- **M24 — `qwp_ws_ownership.rs` is misnamed.** Contains diagnostic/public-API types (`QwpWsProgress`, `QwpWsSenderError`, `QwpWsErrorHandler`, etc.), no ownership types.
- **M25 — Hardcoded `/home/jara/devel/...` paths in committed code.** `qwp_ws_sfa_segment.rs:1190` (under `#[cfg(test)]`, so not production), plus the same path in design docs. Breaks for any other developer not setting `QDB_JAVA_CLIENT_CORE`.
- **M26 — 1,471 lines of "proposal / not implemented" design docs committed alongside the implementation.** `doc/QWP_WEBSOCKET_PROGRESS_CONVERGENCE_DESIGN.md:3`, `doc/QWP_WEBSOCKET_NOTIFICATION_ONLY_ERROR_DELIVERY.md:3`. Either consolidate into the architecture doc or drop them.
- **M27 — `buffer/qwp.rs` is 9,534 lines, mixing two unrelated buffer types and a 4,000-line test module.** Should split `QwpWsColumnarBuffer` into its own file.
- **M28 — `SfaSlotQueue` `.lock.pid` file never cleaned up on shutdown.** `qwp_ws_sfa_slot.rs:241, 305-308`. Leaves stale pid files; future scans see a directory with old pid.

## Test coverage gaps (verified)

- **T1 — Zero positive-path C++ test for QWP/WS.** `cpp_test/test_line_sender.cpp` has 3 QWP/WS tests at lines 2742/2775/2806; all are negative (reject paths). The C++ wrapper's exception translation for `flush`, `published_fsn`, `await_acked_fsn`, `drive_once`, `close_drain` is entirely untested. For a new transport this is a P0 hole.
- **T2 — `line_sender_qwpws_drive_once`** has no Python binding and only a C++ negative test. Manual progress mode is essentially untested cross-FFI.
- **T3 — `line_sender_qwpws_flush_and_keep_and_get_fsn`** has zero callsites outside Rust unit tests.
- **T4 — No loom / multi-threaded property test for the SFA queue.** `grep -rn "cfg(loom)\|loom::"` returns nothing. The queue has Acquire/Release orderings that are not stress-tested.
- **T5 — WSS coverage is one happy-path Rust test (`sender_qwp_wss_round_trip_delivers_replay_payload` at qwp_ws_driver.rs:3870).** No handshake-failure, cert/hostname-validation, or `tls_to_plain_server` analogues.
- **T6 — No FFI test crossing through `line_sender_error_qwpws_get_view`** from a real terminal-error path.
- **T7 — QWP/WS f64-array and decimal columns are not tested.** Zero references in `questdb-rs/src/tests/qwp_ws.rs`.
- **T8 — `QwpWsErrorCategory::SecurityError` / `InternalError` end-to-end paths exist only behind `#[ignore]` probes.**
- **T9 — All real-server probe tests (`qwp_ws_protocol_probe.rs:79` etc.) are `#[ignore]` plus env-var-gated.** `ci/run_tests_pipeline.yaml` does not set the env vars, so they never run in CI.
- **T10 — No FFI test for the `Halt` policy via C++.** Python only exercises `DROP_AND_CONTINUE`.
- **T11 — `Buffer::clone()` cross-buffer bookmark independence not tested at the public Buffer level.**
- **T12 — `multicast_ttl` set is asserted but multicast send/receive is not exercised.**
- **T13 — `from_conf` parser smoke test never validates parsed-state values.**
- **T14 — Java golden parity is one fixed scenario.**

## Minor / Nits

- N1 Dead `SchemaRegistry` type (qwp.rs:3889-3925). Delete.
- N2-N4 Naming churn: `qwp_ws_*` vs `qwpws_*` vs no-prefix on `Sender` methods. Pick one.
- N5 `cargo doc` regresses with 3 broken intra-doc links (`buffer/ilp.rs:163, 261`, `decimal.rs:201`).
- N6 Production `unwrap()` without justification: `qwp_ws.rs:1518`, `qwp_ws_driver.rs:2762`.
- N7 `auth_timeout_millis` / `close_flush_timeout_millis` are conf-string-only; HTTP `request_timeout` is `pub fn`. Asymmetric.
- N8 Mixed `_ms`/`_millis` suffix in QWP/WS conf keys.
- N9 `line_sender_bookmark` is exposed as a non-opaque 2×u64 struct; the rest of the library uses opaque pointers. Future changes become ABI-breaking.
- N10 `line_sender_buffer_new_qwp` (no-arg) and `_new_qwp_with_max_name_len` redundant in C; only the latter exposed in C++.
- N11 SfaAckWatermark `Relaxed` FSN store with `Release` flag; no `msync` for durability.
- N12 `record_all_sfa_cleanup` skips ack-watermark cleanup if segment cleanup fails.
- N13 Quarantine rename failure silently swallowed.
- N14 `line_sender_qwpws_close_drain` does not consume; user must still call `line_sender_close`. Doc could be stronger.
- N15 Background runner threads are anonymous (no `Builder::name`).
- N16 50µs idle sleep in runner loop (burns CPU); Condvar would be better.
- N17 `line_sender_opts_clone` C signature takes non-const `opts*` but Rust takes `*const`.
- N18 Internal doc drift: `multicast_ttl` documented as `u8`, actual is `u32`.
- N19 `compute_accept` non-RFC-compliant key not validated (test-only callers).
- N20 `parse_http_header_block` silently skips no-colon lines.
- N21 `await_acked_fsn(Duration::MAX)` saturates to "wait forever" rather than `Ok(false)`.
- N22 Pre-commit hook exists at `.githooks/pre-commit` but is opt-in (`core.hooksPath` unset); the PR description's claim about enforcement is wrong.
- N23 Several commit titles exceed 50 chars — squash/rewrite before merge.
- N24 `BackpressureNotifier` locking discipline undocumented.
- N25 `validate_submit` non-atomic read of `completed_upper`/`published_upper` — under-counts in safe direction.
- N26 `Buffer` auto-derives `Sync` post-rework; future interior-mutation refactor would silently break the FFI contract.
- N27 `Sender::flush_and_keep_and_get_fsn` is verbose; consider `flush_keep_get_fsn`.
- N28 `pub` constants in `ingress.rs` lack `///` docs (`MAX_ARRAY_BUFFER_SIZE`, `MAX_ARRAY_DIM_LEN`, `DECIMAL_BINARY_FORMAT_TYPE`).
- N29 `#[doc(hidden)] pub fn qwp_ws_terminal_error` — should be `pub(crate)`.
- N30 `SymbolGlobalDict::intern` allocates twice per new symbol (cold start only).
- N31 `column_lookup_key` allocates a fresh `String` on cursor miss despite the column already caching `lower_ascii_name`.
- N32 `QwpWsColumnarBuffer::len()` is O(sum of cells); document or cache a running total.
- N33 Test-only `encode_ws_message` allocates per-call inner `Vec`s while production replay reuses a scratch — code-drift hazard. Remove the dead test-only path.
- N34 QWP/UDP per-datagram syscall; no `sendmmsg` batching opportunity.

## Downgraded (false positives or over-stated)

Highlights: 2^64-collision arguments on bookmark generations, defensive `checked_add` on sequence overflow paths that require >zettabyte budgets, theoretical `AtomicI64::from_ptr` alignment concerns on already-page-aligned mmaps, `compute_accept` security gap on a `pub(super)` test-reachable function with only RFC-compliant production nonces, frame_count torn-read concern in the producer/consumer queue (today self-correcting, retried by the reader), and a handful of pre-existing patterns (SIGPIPE masking, `line_sender_error_free` panic exposure) not introduced by this PR but widened by it. Roughly 38 of the original 180+ candidate findings were dismissed after re-reading source.

## Summary

**Verdict: request changes.** This is an ambitious, mostly well-structured feature, but it ships with a confirmed downstream-binding-breaking ABI change (C1), five FFI panic-unsafety regressions (C2–C5), an unbounded memory leak in idiomatic flush-loop usage (C6), and a Drop-time hang + unbounded background-thread / FD / mmap leak when the server is unreachable (C7). None of these are subtle to fix.

- **180+ draft findings → 70 verified, 38 downgraded as false positives / over-stated.**
- **7 Critical, ~28 Moderate, 14 Test gaps, 34 Minor.**
- **Vast majority in-diff** (most code is new). A handful contract-shifted out-of-diff items (SIGPIPE surface widening, `line_sender_error_free` asymmetry, the orphan-drainer interaction with existing `Drop` ordering).
- **Cross-context exposure check was skipped under the level-2 budget.** Recommend running a level-3 review before merging since the FFI surface is large and the diff touches almost every entry point.

bluestreak01 (approved):

PR #142, opened 2026-05-15
chore(egress): egress fuzz tests

jovfer:
port fuzz tests:
- borrow CI infra from #141 
- port Java fuzz tests from [OSS db main repo](https://github.com/questdb/questdb/tree/master/core/src/test/java/io/questdb/test/cutlass/qwp)

### DIFF
--- a/questdb-rs/tests/egress_live_server_fuzz.rs
+++ b/questdb-rs/tests/egress_live_server_fuzz.rs
@@ -66,9 +66,9 @@
 //!     default / `compression=raw` / `compression=auto` /
 //!     `compression=zstd` / `compression=zstd;compression_level=N`
 //!     with `N ∈ [1, 9]`.
-//!   - `testSelectAlterSequenceFuzz` (ALTER-orchestration fuzz) is in
-//!     its own follow-up commit (separate driver — too different
-//!     structurally to share `run_one_case`).
+//!   - `testSelectAlterSequenceFuzz` (ALTER-orchestration fuzz) lives
+//!     in a sibling file (`egress_live_server_alter_fuzz.rs`); its
+//!     driver is structurally too different to share `run_one_case`.
 //!
 //! Each test boots its own `QuestDbServer` via `start_with_config(...)`.
 //! Three tests × one JVM each ≈ 45 s of boot for the whole file,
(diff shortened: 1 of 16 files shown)


questdb-rs/tests/egress_live_server_fuzz.rs
```diff
@@ -0,0 +1,1261 @@
...
+            }
+        })
+        .unwrap_or(DEFAULT_SEED);
+    let mut hash: u64 = 0xCBF2_9CE4_8422_2325;
+    for b in test_name.bytes() {
+        hash ^= b as u64;
+        hash = hash.wrapping_mul(0x100_0000_01B3);
+    }
+    let combined = base.wrapping_add(hash);
+    eprintln!("[qwp_egress_fuzz seed] {test_name} seed=0x{combined:016x}");
```
jovfer, line 222: The printed `seed=0x{combined}` isn't a useful reproduction key — at runtime `name_hash` is added to whatever is in `QWP_EGRESS_FUZZ_SEED`, so pasting `combined` into the env var or into `DEFAULT_SEED` yields a *different* sequence. The only way to actually replay from this line is to bypass `fuzz_seed_for` and hardcode `SplitMix64::new(0x<combined>)` at the call site.

Compare with the Java pattern this was ported from — `TestUtils.generateRandom(LOG, s0, s1)` prints `random seeds: <s0>L, <s1>L`, and those values are exactly what you paste into `new Rnd(s0, s1)`. **The print is the recipe.**

Minimum fix — also print `base` so the env var is copy-pasteable:

```suggestion
    eprintln!("[qwp_egress_fuzz seed] {test_name} base=0x{base:016x} combined=0x{combined:016x} (reproduce: QWP_EGRESS_FUZZ_SEED=0x{base:016x})");
```

Bigger, separate decision: `DEFAULT_SEED` on line 202 makes local runs deterministic — that's regression-pinning, not fuzz. Java defaults to `nanoTime() + currentTimeMillis()`, so each local invocation explores fresh territory. CI varies via `Build.BuildId` either way, but local `cargo test` always explores the same region.

Same `fuzz_seed_for` exists in `egress_live_server_alter_fuzz.rs`, `egress_live_server_bind_fuzz.rs`, `egress_live_server_fragmentation_fuzz.rs` — leaving one canonical comment here, happy to fan it out if we take the change.

PR #150, opened 2026-05-26
feat: Apache Arrow / Polars integration for QWP-WS

kafka1991:
## Summary

Adds Apache Arrow + Polars integration across both directions of the QuestDB Wire Protocol (WebSocket transport):

- **Egress** (server → client): `Cursor::as_record_batch_reader()` adapter that yields `arrow::RecordBatch` from streaming query results; optional Polars bridge via Arrow C Data Interface.
- **Ingress** (client → server): `Buffer::append_arrow()` writes an entire `RecordBatch` into a QWP-WS buffer via a column-major dense bulk path — primitive / string / binary columns use a single memcpy per column instead of the row-by-row HashMap lookup that the existing per-cell API takes.

## Public surface

**Rust** (`questdb-rs`):

- `Buffer::append_arrow(table, batch)` — ingress; server stamps each row's timestamp on arrival (same semantics as `at_now()`).
- `Buffer::append_arrow_at_column(table, batch, ts_column)` — ingress; sources the per-row designated timestamp from a named `Timestamp(_)` column inside the batch (must have no null rows).
- `Cursor::as_record_batch_reader() -> CursorRecordBatchReader` — egress, streaming `Iterator<Item = Result<RecordBatch, ArrowError>>`.
- `Cursor::fetch_all_arrow() -> (SchemaRef, Vec<RecordBatch>)` — egress, eager drain.
- `Cursor::next_polars() / iter_polars() / fetch_all_polars()` — polars sub-feature.
- `polars::dataframe_to_batches(df, max_rows)` — chunked DataFrame → RecordBatch iterator for ingress.

**C ABI** (gated on `QUESTDB_ENABLE_ARROW`):

- `line_reader_cursor_next_arrow_batch(...)` — egress via Arrow C Data Interface
- `line_sender_buffer_append_arrow(...)` / `line_sender_buffer_append_arrow_at_column(...)` — ingress via Arrow C Data Interface
- `line_sender_buffer_new_qwp_ws()` — construct the QWP/WebSocket columnar buffer the Arrow ingest path expects.
- New error codes: `schema_drift`, `no_schema`, `arrow_export`, `arrow_unsupported_column_kind`, `arrow_ingest`

## Dense bulk path

QWP-WS buffer gains 7 Arrow-fed column variants (`ArrowFixed` / `ArrowVarLen` / `ArrowBool` / `ArrowSymbol` / `ArrowDecimal` / `ArrowGeohash` / `ArrowArray`) that store wire-format-ready bytes directly. `append_arrow` is column-major (one HashMap lookup per column, not per cell) and per-kind correctness follows the QWP sparse-null vs sentinel convention: `i32::MIN` / `f64::NAN` for non-nullable signed integer / float kinds; non-null + bitmap for kinds that support it. `UInt8` widens to `Int32` and `UInt64` is rejected when `value > i64::MAX`, avoiding silent sentinel collisions.

For batches of N=10K rows × M=20 primitive columns with no nulls, this collapses the work from ~200K per-cell HashMap operations to 20 memcpys. With nulls present, each column's QWP null bitmap is built by byte-stride OR-with-NOT of the Arrow validity buffer when source and destination boundaries align; per-row fallback only when bit-offsets are unaligned.

## Feature gating

- `questdb-rs/Cargo.toml`: `arrow` + `polars` features (opt-in, excluded from `almost-all-features`)
- `questdb-rs-ffi/Cargo.toml`: `arrow` feature mirrors
- `CMakeLists.txt`: `QUESTDB_ENABLE_ARROW=OFF` by default; auto-flipped to `ON` when `QUESTDB_TESTS_AND_EXAMPLES=ON` so tests / examples can exercise the Arrow path without explicit opt-in.

## FFI safety

- Schema depth pre-validated (cap 64) before `arrow::ffi::from_ffi` recurses through `children`.
- Top-level Struct with root null bits → `arrow_ingest` error, not an `assert!` abort.
- `row_count` capped at 16 M and per-column `value_data()` at 1 GiB on the FFI entry; bounds every `Vec::reserve` against allocator-OOM aborts under the FFI crate's `panic = "abort"` profile.
- Cursor honours its `terminal_error` replay contract on the Arrow path: a transport error captured on a previous call replays on every subsequent `_next_arrow_batch`, no silent `Ok(None)` truncation.

## Tests

- **Rust unit tests**: 80+ tests in `questdb-rs/src/egress/arrow/tests.rs` + `src/ingress/arrow.rs` covering each Arrow → QWP kind, multi-batch dense extension, sliced array (primitive / utf8 / bool with non-zero offset), Decimal256 round-trip, mid-stream schema drift (tentative-ndim upgrade vs firm-firm drift), UInt64 sentinel rejection, geohash precision width check.
- **C ABI** ([`cpp_test/test_arrow_c.c`](cpp_test/test_arrow_c.c)): enum constants, struct layouts, NULL-safety, per-type dispatch with strict-success assertions (no soft-error masking).
- **C++ doctest**: [`cpp_test/test_arrow_egress.cpp`](cpp_test/test_arrow_egress.cpp) + [`cpp_test/test_arrow_ingress.cpp`](cpp_test/test_arrow_ingress.cpp)
- **System tests** (live QuestDB):
  - `system_test/arrow_egress_fuzz.py` — egress fuzz
  - `system_test/arrow_ingress_fuzz.py` — ingress fuzz
  - `system_test/arrow_round_trip_fuzz.py` — write Arrow → flush → SQL readback → row-level compare
  - `system_test/arrow_alignment_fuzz.py` — alignment regression
- **CI** ([`ci/run_all_tests.py`](ci/run_all_tests.py)): wires the three new cpp_test binaries + runs cargo tests with `arrow,polars` features

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -107,7 +107,7 @@ target_include_directories(
     questdb_client INTERFACE
     ${CMAKE_CURRENT_SOURCE_DIR}/include)
 if(QUESTDB_ENABLE_ARROW)
-    target_compile_definitions(questdb_client INTERFACE QUESTDB_CLIENT_HAS_ARROW)
+    target_compile_definitions(questdb_client INTERFACE QUESTDB_CLIENT_ENABLE_ARROW)
 endif()
 if(WIN32)
     set_target_properties(
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,11 @@ option(
     "Build the C/C++ tests with -fsanitize=address,undefined."
     OFF)
 
+option(
+    QUESTDB_ENABLE_ARROW
+    "Build with Apache Arrow C Data Interface exports. Opt-in: pulls arrow-rs."
+    OFF)
+
 # Build static and dynamic lib written in Rust by invoking `cargo`.
 # Imports `questdb_client` target.
 add_subdirectory(corrosion)
@@ -80,6 +85,13 @@ if(QUESTDB_ENABLE_READER)
 endif()
 if(QUESTDB_ENABLE_INSECURE_SKIP_VERIFY)
     list(APPEND QUESTDB_CARGO_FEATURES insecure-skip-verify)
+endif()
+if(QUESTDB_TESTS_AND_EXAMPLES AND NOT QUESTDB_ENABLE_ARROW)
+    message(STATUS "QUESTDB_TESTS_AND_EXAMPLES=ON: enabling QUESTDB_ENABLE_ARROW")
+    set(QUESTDB_ENABLE_ARROW ON)
+endif()
+if(QUESTDB_ENABLE_ARROW)
+    list(APPEND QUESTDB_CARGO_FEATURES arrow)
 endif()
 if(QUESTDB_CARGO_FEATURES)
     corrosion_import_crate(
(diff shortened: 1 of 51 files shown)


kafka1991:
included in https://github.com/questdb/c-questdb-client/pull/150

PR #152, opened 2026-05-29
chore: center QWP/WS reconnect backoff jitter

jerrinot:
The jitter was additive -- `base + rand(0, base)` -> `[base, 2*base)`, so it never retried sooner than the full backoff. Under rapid server restarts the client sampled too coarsely to catch the brief server-up windows and close_drain timed out in the QWP/WS bounce fuzz. Center it instead: `base/2 + rand(0, base)` -> `[base/2, 3*base/2)`, so retries scatter around the backoff and can fire at half it (`reconnect_max_backoff` is now the jitter center, not a hard cap). 

Also cap the bounce-fuzz producers at 250ms to fix flakiness.

### DIFF
--- a/questdb-rs/src/egress/reader.rs
+++ b/questdb-rs/src/egress/reader.rs
@@ -2892,13 +2892,13 @@ mod tests {
     }
 
     /// Every draw lies in `[0, base)` — the full-jitter contract from
-    /// failover.md §3.1. This is the discriminator against the
-    /// equal-jitter variant `[base, 2*base)` used by SF ingress: a
-    /// regression that swapped the implementation would produce draws
-    /// of `base` or higher on the first iteration. 10k samples per base across
-    /// several bases (powers of two, near-`u32::MAX`, and primes that
-    /// exercise the `% base` reduction) catches both off-by-one and
-    /// signed/unsigned mix-ups.
+    /// failover.md §3.1. SF ingress uses a different scheme (centered
+    /// jitter, `[base/2, 3*base/2)`, in `qwp_ws_driver.rs`); this test
+    /// pins the egress full-jitter contract, which — unlike that — may
+    /// wait near zero. 10k samples per base across several bases
+    /// (powers of two, near-`u32::MAX`, and primes that exercise the
+    /// `% base` reduction) catches both off-by-one and signed/unsigned
+    /// mix-ups.
     #[test]
     fn full_jitter_ms_draws_are_in_range() {
         let mut rng = FailoverRng::new();
--- a/questdb-rs/src/ingress.rs
+++ b/questdb-rs/src/ingress.rs
@@ -1492,7 +1492,8 @@ impl SenderBuilder {
     }
 
     #[cfg(feature = "_sender-qwp-ws")]
-    /// Cap on the per-attempt reconnect delay. Default 5s.
+    /// Cap on the reconnect backoff the retry loop doubles toward; the actual
+    /// per-attempt delay is this value jittered to ~[half, 1.5x]. Default 5s.
     pub fn reconnect_max_backoff(mut self, value: Duration) -> Result<Self> {
         if value.is_zero() {
             return Err(error::fmt!(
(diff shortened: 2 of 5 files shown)


jerrinot:
I have enough to finalize. This is a small, self-contained change and I've traced it end-to-end: the jitter math, both callers (`qwp_ws_driver.rs:242`, `qwp_ws.rs:662`/`2770`), the deadline clamp, the docs, the tests, and the system-test config key. Here is the review.

---

## Reviewing PR #152 at level 0 (default)

`chore: center QWP/WS reconnect backoff jitter`. Single-pass inline review (correctness, panics, FFI, tests, standards). For a change of this size and risk, level 0 is appropriate — but note level 3 exists for a full multi-agent pass if you want one.

### What the change actually does

`equal_jitter_duration` → `centered_jitter_duration`, swapping the additive scheme `base + rand(0, base)` → `[base, 2·base)` for `base/2 + rand(0, base)` → `[base/2, 3·base/2)`. Plus doc updates on `reconnect_max_backoff` (builder + conf), a comment realignment in `egress/reader.rs`, a new range test, and a 250 ms backoff cap in the bounce fuzz producer.

I verified the math, the panic-safety, all three callsites, the deadline interaction, and the test bounds. **No correctness, safety, or FFI issues found.** Details below.

### Critical
None.

### Moderate
None.

### Minor

1. **The new test is vacuous unless `sync-sender-qwp-ws` is enabled** — `qwp_ws_driver.rs:4196`. When that feature is off, `centered_jitter_duration` is the `base`-returning stub (`qwp_ws_driver.rs:1878`), so `reconnect_sleep_duration(false, …, base)` returns exactly `base`, which trivially satisfies `d >= base/2 && d < base + base/2`. The test still passes, but it only *exercises* the jitter when run with the feature flag. Not a bug — worth knowing the test asserts nothing about jitter in a non-sync build. If you want it to fail loudly when run in the wrong config, gate it `#[cfg(feature = "sync-sender-qwp-ws")]`.

2. **`reconnect_max_backoff` is no longer a hard cap on the per-attempt delay** — now the effective sleep at the cap is `[max_backoff/2, 1.5·max_backoff)`. This is a real semantic change to a `pub` builder method (`ingress.rs:1495`), but it is correctly disclosed in both the builder doc and the PR body, and `sleep_before_runner_reconnect` clamps the result to the remaining deadline (`qwp_ws.rs:1323`), so it can never overshoot `reconnect_max_duration`. The docstring's "jittered to ~[half, 1.5x]" is exact only once `backoff` has ramped to the cap; below the cap it's centered on the smaller current backoff. The `~` and "value the retry loop doubles toward" phrasing cover this adequately. No action required.

### Things I checked that are fine
- **Panic safety:** `rand::rng().random_range(0..base_nanos)` would panic on an empty range, but the `base_nanos == 0` early return (`qwp_ws_driver.rs:1865`) guarantees `base_nanos >= 1`. `saturating_add` + the `.min(u64::MAX)` clamp make the `~1.5×` arithmetic overflow-proof. No reachable panic.
- **Caller ordering unchanged:** `record_retryable_error` computes `sleep_for` from the current backoff *then* doubles it (`qwp_ws_driver.rs:241-247`) — same order as before; the async caller (`qwp_ws.rs:662-674`) matches.
- **System test config key valid:** `reconnect_max_backoff_millis` is an existing key (already used at `test.py:2067`); 250 ms is sound and the comment correctly notes it's inert for non-bounce variants.
- **No stale `equal_jitter` references** anywhere in real source (header/cpp/python/docs all clear).
- **`egress/reader.rs` change is comment-only** — the full-jitter `[0, base)` scheme there is untouched; the edit just clarifies it's a *different* scheme from the new centered one.
- **No FFI/C/C++ surface touched**; no new public symbols.
- **The other existing test** (`role_reject_reconnect_sleep_uses_fixed_initial_backoff`, `qwp_ws_driver.rs:4189-4193`) still holds: `role_reject=true` → `initial` unchanged; `ZERO` base → `ZERO` via the early return.

### Downgraded (false positives)
None — no draft findings were dismissed; the inline pass produced no false leads worth recording.

### Summary
**Verdict: approve.** Clean, well-scoped chore with correct jitter math, a guarded no-panic implementation, accurate doc updates across all three sites, and a regression test for the new range. The only behavioral consequence — per-attempt delay reaching up to 1.5× `reconnect_max_backoff` — is intentional, documented, and bounded by the deadline clamp. 0 findings in-diff, 0 out-of-diff (cross-context surface is genuinely tiny: one private jitter fn with two internal callers, both verified). Two minor notes above, neither blocking.

jerrinot:
an unrelated e2e client flake before https://github.com/questdb/questdb-enterprise/pull/1036 was merged

mtopolnik (approved):

bluestreak01 (approved):

PR #148, opened 2026-05-24
feat(ingress): column-major sender for QWP/WebSocket (WS-0..WS-6)

bluestreak01:
## Summary

Lands the Rust core, C ABI, and benchmarks for a column-major sender that
ingests Pandas/Polars DataFrames into QuestDB at the maximum throughput
the QWP/WebSocket wire allows. The design and FFI spec landed in earlier
commits on this branch (`doc/COLUMN_SENDER_PLAN.md`,
`doc/COLUMN_SENDER_FFI_ABI.md`); this PR implements WS-0…WS-6 of the
plan in a single commit.

The user model is `DataFrame → Table`:

```rust
let db = QuestDb::connect("qwpws::addr=localhost:9000;pool_size=4;")?;
let mut sender = db.borrow_sender()?;
let mut chunk = Chunk::new("trades");
chunk.column_f64("price", &prices, None)?;
chunk.column_i64("qty",   &qty,    None)?;
chunk.symbol_dict_i32("sym", &codes, &dict_offsets, &dict_bytes, None)?;
chunk.designated_timestamp_nanos(&ts)?;
sender.flush(&mut chunk, AckLevel::Ok)?;  // returns when server ACKs
```

The same surface is exposed verbatim through the C ABI in
`include/questdb/ingress/column_sender.h`.

## What's in the box

* **WS-0 — `QuestDb` pool** (`ingress/column_sender/{db,conf}.rs`).
  Thread-safe; eager-open, fail-fast at `pool_max`, `BorrowedSender<'a>`
  that returns on `Drop`, background reaper under `pool_reap=auto`.
  New conf keys: `pool_size`, `pool_max`, `pool_idle_timeout_ms`,
  `pool_reap`. Store-and-forward (`sf_*` / `sender_id`) and
  `qwp_ws_progress=manual` refused at `connect`-time.

* **WS-1 — synchronous `flush`** (`ingress/column_sender/sender.rs`,
  `encoder.rs`). `ColumnSender::flush(chunk, AckLevel)` encodes, publishes
  via the existing QWP/WS replay queue, and blocks until the ACK
  watermark crosses the published FSN. Polls in 50 ms slices so
  `must_close` mid-wait surfaces promptly. `AckLevel::Durable` requires
  `request_durable_ack=on` at connect (else `InvalidApiCall`).

* **WS-2 — `Chunk` + numeric / fixed-width columns**
  (`chunk.rs`, `validity.rs`, `wire.rs`).
[...]

PR #153, opened 2026-06-04
feat: columnar DataFrame ingest (Arrow / Polars / Pandas) + Arrow egress

kafka1991:
## Summary

This PR is the combined landing of **#148** and **#150** — two independently-developed columnar I/O tracks that both target QWP/WebSocket and share the same connection pool / `SymbolGlobalDict`, shipped together so downstream consumers (`py-questdb-client`, in-tree C/C++ tests) see one self-consistent surface.

### #148 — Column-major sender (`column_sender`)

A DataFrame → Table ingest API. `QuestDb` connection pool + `BorrowedSender` + `Chunk` (per-column `Vec<u8>` that stacks wire bytes directly) + synchronous `flush(AckLevel)`. Covers bool / signed integers / floats / UUID / Long256 / IPv4 / timestamps / VARCHAR / `symbol_dict_{i8,i16,i32}` bulk-intern. The connection-scoped `SchemaRegistry` (FULL / REFERENCE emit modes) and `SymbolGlobalDict` are shared with the row API, preserving the 1M-per-connection symbol cap on huge Pandas `Categorical` dicts. Full C ABI (`include/questdb/ingress/column_sender.h`) and a Criterion bench suite (column path ≈ memcpy ceiling; bulk-intern ~16× faster than per-row HashMap).

### #150 — Apache Arrow + Polars integration

Both directions over QWP/WebSocket:

- **Ingress**: `Buffer::append_arrow` / `append_arrow_at_column` consumes a whole `RecordBatch` in one call, column-major dense bulk path (one memcpy per column; QWP null bitmap built by byte-stride OR-with-NOT of the Arrow validity buffer when boundaries align, per-row fallback only when bit-offsets are unaligned).
- **Egress**: `Cursor::as_record_batch_reader()` streaming `RecordBatch` iterator; Polars sub-feature provides the DataFrame bridge.
- **C ABI** via the Arrow C Data Interface: `line_sender_buffer_append_arrow*` and `line_reader_cursor_next_arrow_batch`. Every producer-supplied `ArrowArray`/`ArrowSchema` is pre-validated before `from_ffi` (schema depth ≤ 64, `row_count` ≤ 16M, bounded per-node buffer/child counts, and rejection of NULL or under-sized buffer arrays), so a malformed struct returns an error instead of aborting the FFI crate's `panic = "abort"` profile. The manual per-column FFI path caps each variable-length payload at `i32::MAX`.

### Why merged

The two tracks were developed on the same `jh_conn_pool_refactor` branch and converged on shared infrastructure (connection pool, `SymbolGlobalDict`, QWP/WS transport). Splitting them at review time would force one to ship behind a compatibility shim for the other; merging them together avoids that churn and gives C/C++ callers — and the upcoming Pandas / Polars wrapper in `py-questdb-client` — a column-major, zero-redundant-copy path into QuestDB in one cut.

## Public surface

See the original PRs' "Public surface" / "What's in the box" sections:

- column_sender API: #148
- Arrow / Polars API: #150

## Feature gating

- `questdb-rs`: `arrow` + `polars` are opt-in features, excluded from `almost-all-features`; column_sender lives behind `sync-sender-qwp-ws`.
- `questdb-rs-ffi`: `arrow` feature mirrors.
- `CMakeLists.txt`: `QUESTDB_ENABLE_ARROW=OFF` by default; auto-flipped to `ON` when `QUESTDB_TESTS_AND_EXAMPLES=ON` so tests / examples exercise the Arrow path without explicit opt-in.

## Test plan

- [x] Rust unit tests: 57 column_sender + 80+ Arrow
- [x] FFI unit tests: 8 new column_sender + Arrow path coverage
- [x] C/C++ tests: `test_arrow_c.c` / `test_arrow_egress.cpp` / `test_arrow_ingress.cpp` wired into CMake and exercised in CI
- [x] System tests against a live QuestDB: `arrow_egress_fuzz` / `arrow_ingress_fuzz` / `arrow_round_trip_fuzz` / `arrow_alignment_fuzz`
- [x] `cargo bench --features sync-sender-qwp-ws --bench column_sender`
- [x] End-to-end Pandas / Polars throughput (`py-questdb-client`, WS-7)

Closes #148, #150.

## Summary by CodeRabbit

* **New Features**
  * Opt-in Apache Arrow support for Arrow egress and ingest; new column-major sender for high-throughput DataFrame ingestion; Polars integration.

* **Examples**
  * Added C, C++ and Rust examples demonstrating Arrow egress/ingest and column-sender/Polars workflows.

* **Documentation**
  * Added column-sender ABI spec, implementation plan, and performance guidance.

* **Tests**
  * Expanded Arrow/Polars unit, smoke and fuzz coverage; new integration tests.

* **Chores (CI)**
  * CI updated to install pyarrow/polars, expanded test/fuzz jobs and longer timeouts.

* **Benchmarks**
  * New Criterion benchmarks measuring column-sender performance.


bluestreak01:
# Review — columnar DataFrame ingest (Arrow / Polars / Pandas) + Arrow egress

**Adversarial level-3 review.** Scope: 90 files, +40,097 / −972 — the combined landing of #148 (column_sender) + #150 (Arrow/Polars).

> **Disclosure:** this review was produced with significant automated assistance, and the reviewer authored part of this PR (the `borrow_row_sender` pool, `PolarsIngestOptions`, `impl TryInto<TableName>`). Findings on those areas are effectively self-review — weight an independent pass on the pool/polars/ergonomics accordingly.

**Method:** build-profile facts captured; deep passes over the ~15 risk-bearing Rust/FFI files; docs/CI/examples/python-fuzz/bench treated as lower priority. Every finding below was verified against source (including arrow-rs 58.3.0).

**Premise (load-bearing):** `questdb-rs-ffi` is `panic = "abort"` in both release and dev (`questdb-rs-ffi/Cargo.toml:72-76`) → `catch_unwind`/`panic_guard` are no-ops; every reachable panic from an FFI entrypoint is a host-process abort.

**Headline:** unusually well-hardened. The highest-risk surface — the Arrow C Data Interface pre-`from_ffi` validation under panic=abort — was verified to close every eager-deref/assert/underflow site arrow-rs actually has. **No confirmed memory-unsafety or data-corruption bug is reachable on the release profile through the documented, validated FFI path.** The findings are an ergonomics data-loss footgun, a debug-only abort, FFI hardening gaps, and consistency/resource issues.

---

## Critical (request changes)

**C1 — `flush()` defers commit; dropping the sender (esp. the C++ RAII destructor) silently discards uncommitted rows. Easy path ≠ safe path.** *(verified)*
- `ColumnSender::flush` sends `FLAG_DEFER_COMMIT` and does not wait for an ack (`sender.rs:164-170`). Drop only poisons the conn — `db.rs:829-832` / `868-871`: `if sender.conn.in_flight() > 0 { mark_must_close() } return_to_pool(...)` — it **never syncs**. So `flush` + drop-without-`sync()` silently drops every uncommitted frame.
- `first_frame_sent` travels with the pooled slot, so a *recycled warm* connection defers **even the first** flush. Identical user code loses everything on a recycled conn but commits on a cold one — nondeterministic from pool state.
- C/C++ is the dangerous surface: `column_sender.hpp` `~borrowed_conn() noexcept { release(); }` drops silently; the C header's flush/sync docs never state "un-synced rows are not committed." Row-API `Sender::flush` (HTTP) commits on return — the *same verb* has the opposite durability contract.
- **Fix:** best-effort `sync` on Drop (or a loud `debug_assert`/log when `in_flight()>0`), and prominently document "MUST call `sync`/`column_sender_sync` or lose uncommitted rows" on `flush`, the C `column_sender_flush`, and the `borrowed_conn` dtor. *(Caveat: documented on Rust `flush` and arguably intended pipelining design — flagging for an explicit decision, not asserting an accidental bug.)*

---

## Moderate

**M1 — `column_sender_arrow_import_new` takes `symbol_mode` as a by-value `#[repr(u32)]` enum → out-of-range C value is instant UB.** *(verified)* The file's own comment (`column_sender.rs:273`) says `dtype`/`ack_level` are taken as `u32` *specifically* to make out-of-range a recoverable `InvalidApiCall` instead of UB — but `column_sender_arrow_import_new` (`:1348`) takes `symbol_mode: column_sender_symbol_mode` by value; a C caller passing `5` materializes an invalid discriminant (UB) before the `match` at `:1352`. **Fix:** make it `u32` + map with `InvalidApiCall` on out-of-range, mirroring `ack_level_from_u32`.

**M2 — numpy append has a data pointer + row_count but no buffer byte-length → a wrapper bug = OOB read (host crash/UB).** *(verified)* `column_sender_chunk_append_numpy_column(data: *const u8, row_count, dtype, ...)` (`column_sender.rs:2016`) reads `row_count * elem_size(dtype)` bytes trusting caller dtype/count; nothing carries `nbytes`. An `int8` array tagged `f64`/`geohash_i64` reads 8× the buffer. Unlike the Arrow path (validated via `from_ffi` pre-checks), the numpy path has no length to even sanity-check. **Fix:** thread a `data_len`/`nbytes` param and assert `row_count.checked_mul(elem_size) <= data_len` before any read; reject non-C-contiguous / negative-stride in the wrapper.

**M3 — Negative `FixedSizeList` size bypasses the Arrow validator → debug-build abort.** *(verified)* `validate_arrow_array_depth` guards negative `FixedSizeBinary(width)` (`lib.rs:4058-4064`) but not `FixedSizeList(_, size)`. A schema `+w:-1` with `n_buffers=2` passes the floor/cap checks, then arrow-rs `bit_width` computes `child_bits * (-1 as usize)`. Under dev `panic=abort` + overflow-checks → abort; release wraps and `validate_full` then rejects gracefully (so: **debug-only abort, release-safe**). No `+w:-N` test exists. **Fix:** also reject negative `FixedSizeList` size (mirror the FixedSizeBinary guard).

**M4 — Pool `health` Mutex held across the blocking connect handshake.** *(verified)* `connect_conn_pool` (`db.rs:1163`) locks `inner.health`, then `connect_conn` → `connector.connect_round(&mut health)` does full TCP+TLS+WS-upgrade I/O under the lock. Serializes all concurrent connects pool-wide and head-of-line-blocks transport-dead returns (which also `lock_health`) behind a slow/black-holed connect — degrading the failover path the pool exists for. No deadlock. **Fix:** choose the endpoint under the lock, release it during network I/O, re-acquire to record the result.

**M5 — Egress: per-column degenerate-list node budget (16M) is reset per column, not per batch → OOM/abort amplification.** *(confirmed, reachability nuance)* `MAX_DEGENERATE_LIST_NODES=16M` with `node_budget` recomputed each `array_column_to_arrow` call (`convert.rs:644`); a frame may carry `MAX_COLUMNS_PER_TABLE=2048` array columns, each a `[16M,0]` row → multi-GB retained → OOM (=abort under FFI). Exact per-column wire cost not fully traced. **Fix (regardless):** make the node budget batch-global, or cap array-column count / cumulative expanded nodes per batch.

**M6 — Egress: mid-query decode errors are now failover-eligible → deterministic-corruption reconnect/replay loop + per-round `eprintln!`.** *(confirmed, nuance)* `reader.rs` routes decode `ProtocolError` through `failover_after_stream_failure`; `warn_on_protocol_error_failover` logs each round (`reader.rs:541`). A server deterministically emitting a corrupt frame now reconnect→replay→re-corrupt until the per-Execute budget drains, with unbounded stderr, where the old path failed fast. **Fix:** keep deterministic decode codes terminal (or give decode-driven replays a small dedicated cap); rate-limit the log.

**M7 — Within the column API, table/column names use two different shapes + validation timings.** *(verified)* `column_sender_chunk_new` takes raw `const char* table_name` / `Chunk::new(impl Into<String>)` validated **lazily at flush** (`chunk.rs:423`); `column_sender_flush_arrow_batch` takes a validated `line_sender_table_name` **eagerly** (`column_sender.h:1081`). A pre-validated `line_sender_table_name` can't feed the chunk path, and the same bad name fails at different points. **Fix:** accept the validated newtypes on the chunk/column entrypoints too, or document the split loudly.

**M8 — Designated-timestamp requirement diverges between publish paths.** *(verified)* The chunk path hard-errors without a designated timestamp (`encoder.rs:108`); `flush_arrow_batch` silently server-stamps unless `_at_column` is used — silently ignoring a timestamp column the caller intended as designated (clock-skew / wrong partition), no error. **Fix:** require an explicit server-stamp opt-in on `flush_arrow_batch`, or document the divergence at both sites.

**M9 — `durable_watermarks` is never pruned (`conn.rs:116`, populated `:476`).** Only `pending_durable_targets` is pruned on `sync`. With `request_durable_ack=on`, a server emitting many distinct table names grows the map for the connection's pooled life → slow OOM. **Fix:** prune/cap `durable_watermarks`.

**M10 — No test coverage for sliced / non-zero-offset Arrow arrays — the classic encoder trap.** *(verified by grep)* No `.slice(` anywhere in `arrow_batch.rs` tests; every test builds offset-0 arrays. Sliced arrays reach this encoder via the polars path (after `df.slice`/filter) and FFI. The slice handling reads correct in all paths traced, but a mission-critical column-major unsafe path with zero offset-slice tests is a real gap. **Fix:** add encode tests over `array.slice(1, n)` per `ColumnKind`, including an `offset%8 != 0` bitmap case.

---

## Minor

- **Encoder never calls `validate_full` itself** (`arrow_batch.rs:3205` only checks `null_count ≤ len`); safety is one missing call away for any future native caller of `flush_arrow_batch` passing an unvalidated/`build_unchecked` array.
- **numpy `NumpyDtype::validate()` doesn't range-check decimal `scale`** (`numpy_wire.rs:245`) despite the module doc claiming the caller validates; FFI path is covered, direct Rust callers aren't.
- **numpy bulk paths use panicking `out.reserve(..)`** while only the ndarray path uses `try_reserve`; mitigated upstream but inconsistent under panic=abort.
- **`drifted_batch` not cleared on the raw↔arrow cursor API switch** (`egress.rs:2566`) → stale parked batch replayed out of order on a supported mixed-use path.
- **`symbol_array` assumes `codes.len()==row_count` unguarded** (`convert.rs:507`) and `reader.rs:1658` `.expect("HaveBatch implies last_batch")` → abort-on-invariant-violation under FFI; return `ProtocolError` instead.
- **Pre-existing (not this PR):** `line_sender_utf8_assert`/`table_name_assert`/`column_name_assert` `panic!` on caller input; `line_sender_utf8_init` does `from_raw_parts(buf,len)` with no NULL guard.
- **Redundant double dict-rollback** (`encoder.rs:146/177` + `sender.rs:303/354`) — idempotent today, fragile if `rollback` ever becomes non-idempotent.
- **Strict trailing-byte rejection** (`conn.rs:931`) latches the conn on any extra bytes → breaks forward-compat with a server that appends response fields.

---

## Downgraded (initially suspected, verified safe — high-signal)

- **Arrow FFI is not an abort vector on release for structural malformation.** The pre-`from_ffi` walk (`lib.rs:3950-4136`) closes every arrow-rs eager-deref/assert/underflow site, uses cycle-safe iterative depth/`try_reserve` walks, and rejects negative/over-`MAX_CHUNK_ROWS` length/offset. Verified against arrow-rs 58.3.0. *(Residual truth: buffer byte-size/alignment lies are inherently untrustable across the C Data Interface — worth documenting.)*
- **Null-bitmap construction is bit-exact and in-bounds for all offset/length/slice combinations** (`arrow_batch.rs:639`, both aligned and unaligned byte-stride branches) — the PR's headline risk, verified.
- **numpy uses `read_unaligned` everywhere** for multibyte reads; unaligned numpy buffers do not cause alignment UB.
- **conn.rs parser cannot be panicked by server bytes** — every `try_into().unwrap()` is preceded by a length check; frames capped at 256 MiB; `try_reserve` before `resize`. Flagged `panic!`/`expect` are test-only.
- **In-flight/ack-slot accounting has no off-by-one; no cross-borrower commit** — recycled conns are latched `must_close` when `in_flight()>0`; status `0x00` = committed.
- **Ownership/free across the Arrow CDI is correct** — consume-on-success, double-import rejected (`release.is_none()`), exported batches own refcounted buffers (no use-after-free when the cursor drops), FailoverRetry re-exports a valid `release`.
- **ABI evolvability is sound** — new error codes appended (no mid-enum insertion); `ErrorCode`/`NumpyDtype`/`ArrowColumnOverride`/`AckLevel` are `#[non_exhaustive]`; C++ `noexcept` forwarders call only non-throwing C fns. C↔Rust signature parity verified across ~40 column-sender entrypoints + the 43-value numpy-dtype enum.

---

## Summary

- **Verdict: request changes** — driven by **C1** (silent data-loss footgun on the C/C++ RAII path) and FFI hardening gaps **M1/M2/M3**. The core unsafe machinery (Arrow validation, null-bitmap, wire encoders, pool accounting) is verified sound on the release profile; this is a high-quality PR.
- **Findings:** 1 Critical, 10 Moderate, ~8 Minor; ~8 suspected issues verified-safe and downgraded.
- **Most actionable:** (1) decide C1 (Drop-sync or loud warning + docs); (2) fix M1 (`symbol_mode` → `u32`) and M3 (negative `FixedSizeList`) — small, clearly correct; (3) thread a length into the numpy ABI (M2); (4) add sliced-array encoder tests (M10).
- **Not deeply covered** (time-boxed): the Python system-test fuzzers, C/C++ doctests, CMake/CI wiring, benches, and `doc/*.md` specs — recommend a focused pass on `system_test/arrow_*_fuzz.py` and `cpp_test/test_arrow_*` before merge for CI-parity assurance.

jovfer:
## QWP DataFrame benchmarks — current numbers (pandas vs polars)

Re-ran the **S1-narrow** DataFrame benches **sequentially, one client at a time** against a single local QuestDB (loopback), so there's no cross-run CPU/IO contention in the numbers.

**Setup**
- Schema **S1-narrow** — 5 cols: `ts` TIMESTAMP, `id` LONG, `price` DOUBLE, `sym` SYMBOL (card 8), `note` VARCHAR (len 16); `TIMESTAMP(ts) PARTITION BY HOUR WAL DEDUP UPSERT KEYS(ts)`; **10M rows**.
- Server: QuestDB `9.4.3-SNAPSHOT @ 85aecd600b` (QWP schema 0x3), loopback, Apple Silicon (14 cores).
- Clients: py-questdb-client `9ab8b48` (vendoring c-questdb-client `352f0180`) · c-questdb-client `bb7e721`.
- 5 timed iterations / 2 warmups, **median** reported. Both egress paths read the **same** 10M table.
- Byte rates are on the **on-wire QWP payload** — ingress 450 MB (45.0 B/row), egress 371 MB (37.1 B/row) — which is identical for both clients, so MiB/s is directly comparable.¹

### Ingress — DataFrame → QuestDB

| Stage | Python (pandas) | Rust (polars) |
|---|---|---|
| **e2e live** | **15.01 M rows/s** · 644 MiB/s · 0.63 GiB/s | **13.21 M rows/s** · 567 MiB/s · 0.55 GiB/s |
| CPU floor (no network) | 285 M rows/s · 12.0 GiB/s *(populate)* | 445 M rows/s · 18.6 GiB/s *(encode)* |

e2e ingress is **wire/WAL-bound** (~0.6 GiB/s): the column-encode/populate CPU floors are 20–34× faster, so the bottleneck is the network round-trip + WAL commit, not column serialization. The two clients are within noise of each other.

### Egress — QuestDB → DataFrame

| Path / stage | Python (pandas) | Rust (polars) |
|---|---|---|
| decode floor (no DataFrame) | 69.5 M rows/s · 2461 MiB/s · 2.40 GiB/s | 65.9 M rows/s · 2332 MiB/s · 2.28 GiB/s |
| → **polars** DataFrame | `to_polars` **46.4 M/s** · 1642 MiB/s · 1.60 GiB/s | `fetch_all_polars` **62.0 M/s** · 2195 MiB/s · 2.14 GiB/s |
| → polars, lazy/per-batch | — | `iter_polars` 61.9 M/s · 2193 MiB/s · 2.14 GiB/s |
| → **pandas** (numpy copy) | `to_pandas` **16.0 M/s** · 567 MiB/s · 0.55 GiB/s | — |
| Arrow C-stream (zero-copy) | `__arrow_c_stream__` 40.0 M/s · 1417 MiB/s · 1.38 GiB/s | — |

**Highlights**
- ✅ **Egress fix `bb7e721` confirmed live.** `fetch_all_polars` / `iter_polars` now materialise **10M rows across 611 RESULT_BATCH frames** — the multi-batch SYMBOL case that previously failed with *"Categories name mismatch"*. At 62.0 M/s they land within ~6% of the decode floor (65.9 M/s), so the per-batch vstack is essentially free.
- Rust polars materialise (62 M/s) is **~34% faster** than Python `to_polars` (46 M/s), while the decode floors are ~equal (66–70 M/s) — the gap is DataFrame assembly, not decode.
- Python `to_pandas` (numpy) is **~2.9× slower** than `to_polars` (16 vs 46 M/s): the numpy path bulk-copies each fixed-width column out of the recycled wire buffer. The genuine zero-copy egress surface is the **Arrow batch** (`__arrow_c_stream__` / `iter_arrow`), not numpy `to_pandas`.

<details><summary>¹ byte-basis note & raw medians</summary>

Byte rates use the on-wire QWP payload for **both** clients (ingress `450,000,140 B`; egress `371,283,602 B` for 10M rows). The Python egress harness natively reports the decoded Arrow-table `nbytes` (~480 MB) as its wire proxy, which gives higher MiB/s for the same rows/s (e.g. `to_polars` 2123 vs 1642 MiB/s on-wire). **rows/s is unaffected** and is the apples-to-apples metric.

Raw medians (rows/s):

| direction · path | py-pandas | rust-polars |
|---|---|---|
| ingress · e2e | 15,013,844 | 13,211,622 |
| ingress · CPU floor | 285,434,625 (populate) | 444,770,610 (encode) |
| egress · decode floor | 69,499,063 | 65,858,979 |
| egress · → polars | 46,366,911 | 61,985,728 |
| egress · → polars (iter) | — | 61,934,125 |
| egress · → pandas | 16,017,905 | — |
| egress · arrow-c-stream | 40,021,438 | — |
</details>

jovfer:
## S2-wide (15-col) DataFrame parity — 10M rows

Follow-up to the S1-narrow parity above, now on the **S2 wide** scenario (the Go `qwp-egress-read-wide` anchor): S1-narrow + 5 `DOUBLE` (`d1..d5`) + 5 high-cardinality `SYMBOL` (`s1..s5`, 100k distinct each → 15 cols). Both clients build from one shared schema/value contract (`examples/bench_schema/mod.rs`, commit af91231) so they put byte-identical column data on the wire.

**Methodology** (same as S1): one shared QuestDB (`9.4.3-SNAPSHOT` @ `85aecd600b`, `http.receive.buffer.size=16M`), single-stream, sequential — Rust ingress populates the DEDUP table (`count()==10M` gate), then Rust egress + Python egress read that same table, Python ingress re-populates last. Apple-Silicon loopback. `GiB/s = rows/s × canonical on-wire bytes/row` (ingress 99.67, egress 92.46) because the per-client `wire_bytes` basis differs (py records decoded-Arrow nbytes, rust on-wire), so native `MiB/s` is not cross-client comparable — `rows/s` is.

### Ingress — 10M rows
| role | py-pandas | rust-polars |
|---|---|---|
| CPU floor | 73.8 M/s · 6.85 GiB/s | 59.2 M/s · 5.49 GiB/s |
| e2e DataFrame→wire | 3.4 M/s · 0.32 GiB/s | 2.8 M/s · 0.26 GiB/s |

### Egress — 10M rows
| role | py-pandas | rust-polars |
|---|---|---|
| decode floor | 14.7 M/s · 1.27 GiB/s | 16.6 M/s · 1.43 GiB/s |
| → polars | 6.9 M/s · 0.59 GiB/s | **0.3 M/s · 0.03 GiB/s** ⚠️ |
| → polars (lazy/iter) | — | 0.3 M/s · 0.03 GiB/s ⚠️ |
| → pandas (Arrow) | 13.1 M/s · 1.13 GiB/s | — |
| → pandas (numpy) | 8.9 M/s · 0.77 GiB/s | — |
| → pandas (lazy/iter) | 9.6 M/s · 0.83 GiB/s | — |
| → pandas (nullable) | 7.3 M/s · 0.63 GiB/s | — |
| Arrow C-stream | 5.5 M/s · 0.47 GiB/s | — |

### Takeaways
- **Ingress** is wire/WAL-bound at ~3 M/s @ ~0.3 GiB/s on both clients — the same ceiling as S1, with the 5×100k symbol dict roughly halving GiB/s (0.6 → 0.3) for the same row count.
- **Egress decode** is healthy on both; Rust's decode floor (16.6 M/s) slightly edges Python's (14.7 M/s).
- ⚠️ **`fetch_all_polars` / `iter_polars` collapse to ~0.3 M/s (≈33 s for 10M)** on the high-cardinality symbols — ~55× slower than Rust's *own* decode floor, and ~20× slower than Python's `to_polars`. Decode is fast, so the cost is entirely in the **per-batch polars categorical assembly**: across the 611 `RESULT_BATCH` frames the 5×100k-category dictionaries appear to be rebuilt/remapped every batch (O(batches × cardinality)). This is the Rust analog of the pandas `iter_pandas` high-card regression fixed earlier in `egress.pxi` — invisible at S1 (card 8) and only detonating at high cardinality. Likely site: `questdb-rs/src/egress/arrow/polars.rs`. Python's `fetch_all` is immune because `from_batches → from_arrow` unifies the dictionary once.
- Python's egress winner remains **Arrow-backed `to_pandas`** (13.1 M/s), consistent with S1.

Harness: `SCHEMA=s2-wide` in `examples/qwp_{ingress,egress}_polars.rs`; Python `--schema s2-wide`. Aggregated with `doc/bench_parity_aggregate.py` (`--ingress-bpr 99.6744 --egress-bpr 92.4641`).

jovfer:
## S2-wide 10M egress — re-measured after the symbol-dict egress fixes

Follow-up to the **S2-wide (15-col) DataFrame parity — 10M rows** comment above. Re-running the same 10M S2-wide scenario now that the high-cardinality `SYMBOL` egress fixes have landed on both clients. **The Rust `fetch_all_polars` / `iter_polars` pathology flagged there (≈0.3 M/s, ~33 s) is resolved.**

Same methodology: one shared QuestDB (`9.4.3-SNAPSHOT` @ `85aecd600b`, `http.receive.buffer.size=16M`), single-stream, strictly sequential — Rust ingress populates the DEDUP table (`count()==10M` gate), Rust egress + Python egress read that same table, Python ingress last. Apple-Silicon loopback. `GiB/s = rows/s × canonical on-wire bytes/row` (ingress 99.6744, egress 92.4641); native `MiB/s` is not cross-client comparable on egress (py records decoded-Arrow nbytes, rust on-wire), so `rows/s` is the basis.

**Commits under test** — c-questdb-client `a4e6ba9` (`d85b498` polars `SymbolRegistry` + `b9ef9fe` `SymbolValuesCache`); py-questdb-client `ed22a2f` (`b610640` `_PolarsSymbolRegistry`/`iter_polars` + `ae6fee7` pandas dtype-reuse & `validate=False` + `ed22a2f` `string_view` C-ABI fix). Env: pandas 2.3.3, pyarrow 21.0.0, numpy 2.0.2, Python 3.9 arm64; questdb-rs 7.0.0.

### Egress — 10M rows
| role | py-pandas | rust-polars |
|---|---|---|
| decode floor | 17.5 M/s · 1.51 GiB/s | 16.7 M/s · 1.44 GiB/s |
| → polars | 11.8 M/s · 1.02 GiB/s | **15.2 M/s · 1.31 GiB/s** |
| → polars (lazy/iter) | — | **15.5 M/s · 1.33 GiB/s** |
| → pandas (numpy) | 11.3 M/s · 0.98 GiB/s | — |
| → pandas (Arrow) | 16.4 M/s · 1.42 GiB/s | — |
| → pandas (nullable) | 1.0 M/s · 0.09 GiB/s ⚠️ | — |
| → pandas (lazy/iter) | 9.8 M/s · 0.84 GiB/s | — |
| Arrow C-stream | 1.8 M/s · 0.15 GiB/s ⚠️ | — |

### Ingress — 10M rows (unchanged; wire/WAL-bound)
| role | py-pandas | rust-polars |
|---|---|---|
| CPU floor | 71.3 M/s · 6.62 GiB/s | 56.3 M/s · 5.23 GiB/s |
| e2e DataFrame→wire | 3.4 M/s · 0.32 GiB/s | 3.2 M/s · 0.30 GiB/s |

### Before → after (egress)
| path | before | after | |
|---|---|---|---|
| rust → polars | 0.30 M/s | **15.2 M/s** | ✅ ≈50× — now at the decode floor |
| rust → polars (iter) | 0.30 M/s | **15.5 M/s** | ✅ ≈51× |
| py → polars | 6.9 M/s | 11.8 M/s | ✅ +71% |
| py → pandas (numpy) | 8.9 M/s | 11.3 M/s | ✅ +27% |
| py → pandas (Arrow) | 13.1 M/s | 16.4 M/s | ✅ +25% |
| py → pandas (nullable) | 7.3 M/s | 1.0 M/s | ⚠️ −86% |
| py Arrow C-stream | 5.5 M/s | 1.8 M/s | ⚠️ −68% |

### Takeaways
- ✅ **The Rust →polars pathology is gone.** `SymbolRegistry` (intern the connection dict once, wire-code == physical categorical code, no per-batch remap) takes `fetch_all_polars`/`iter_polars` from ~0.3 M/s to ~15.3 M/s — within 8% of the decode floor (16.7 M/s).
- ✅ **All registry-aware Python paths improved**: `to_polars` +71%, `to_pandas` numpy +27%, `to_pandas` Arrow +25%.
- ⚠️ **Two Python paths regressed: `to_pandas(dtype_backend="numpy_nullable")` (−86%) and the pyarrow-free `__arrow_c_stream__` → `polars.from_arrow` (−68%).** These are the only two egress paths that *don't* go through the new fast-assembly: `numpy_nullable` runs `to_arrow()` → `pyarrow.Table.to_pandas(types_mapper=numpy_nullable)` — the same route as the *improved* `to-pandas-arrow`, differing only in the mapper — and Arrow-C-stream hands the raw capsule to `polars.from_arrow`. Both re-unify the per-batch symbol dictionaries in a generic consumer, i.e. the same O(batches × cardinality) shape the original pathology had, now surfacing on the consumer side (the nullable path burns ~52 s of CPU for 10M×15). The benchmark harness is unchanged between the two runs (the three commits touch only `egress.pxi`), so this is a real runtime change — though a pyarrow/pandas version contribution can't be fully excluded (the prior run's library versions weren't recorded). Worth a look, especially as `__arrow_c_stream__` is the advertised no-pyarrow zero-copy path; an A/B on `egress.pxi`@`b610640~1` (same box/libs, egress only) would attribute it cleanly.

Harness: `SCHEMA=s2-wide` in `examples/qwp_{ingress,egress}_polars.rs`; Python `--schema s2-wide`. Aggregated with `doc/bench_parity_aggregate.py --ingress-bpr 99.6744 --egress-bpr 92.4641`.

jovfer:
## S2-wide 10M egress — both regressions resolved (re-measured)

Follow-up to the [re-measured-after-the-egress-fixes comment](https://github.com/questdb/c-questdb-client/pull/153#issuecomment-4778733688) above, which flagged two Python egress paths that had regressed: `to_pandas(dtype_backend="numpy_nullable")` (−86%) and the pyarrow-free `__arrow_c_stream__` → `polars.from_arrow` (−68%). **Both are now fixed** by the compact-dict / query-scope-dict work that has since landed on both clients.

**Commits under test** — c-questdb-client **`1d88fc8`** (`565f45b` symbol *compact* Arrow ABI, `b232afd` query-scope-dict server flag, `1d88fc8` SYMBOL-cache-after-failover fix); py-questdb-client **`2965b50`** (`491d214` `to_arrow`/`__arrow_c_stream__` use compact dicts, `18b8207`/`f760358` recreate/last-chunk dicts). The py submodule pins the same `1d88fc8`, and the standalone Rust examples build from it too — same core on both sides. **Env identical to the regressed run** (pandas 2.3.3, pyarrow 21.0.0, numpy 2.0.2, polars 1.36.1, Python 3.9 arm64), so the change is attributable to the code, not a library bump.

### The two paths, before → after
| path | original | regressed (prev comment) | **now** | |
|---|---|---|---|---|
| `to_pandas(numpy_nullable)` | 7.3 | 1.0 ⚠️ | **6.7** | ✅ |
| `__arrow_c_stream__` → polars | 5.5 | 1.8 ⚠️ | **6.2** | ✅ |

Two *different* fixes, both **on by default** (the benchmark sets no flag):
- **`numpy_nullable`** rebases every RESULT_BATCH chunk onto the last chunk's (append-only superset) dictionary, so all 611 chunks share one physical dictionary and pandas' compiled `table_to_blocks` dictionary-unify becomes a no-op.
- **`__arrow_c_stream__`** now pulls **compact** batches (`reader_cursor_next_arrow_batch_compact`): each per-batch dictionary carries only the values that batch actually references, so `polars.from_arrow` still unifies 611 dicts but each is minimal. (That's why it recovers yet stays the slowest egress conversion path — the unify is reduced, not eliminated.)

The query-scope flag (`reset_symbol_dict`, default on) is orthogonal here — it resets the connection dict at *query start* to bound cross-query bloat, not the within-query per-batch dicts.

### Egress — 10M rows
| role | py-pandas | rust-polars |
|---|---|---|
| decode floor | 13.3 M/s · 1.14 GiB/s | 13.9 M/s · 1.20 GiB/s |
| → polars | 10.3 M/s · 0.89 GiB/s | 12.4 M/s · 1.07 GiB/s |
| → polars (lazy/iter) | — | 13.2 M/s · 1.14 GiB/s |
| → pandas (numpy) | 10.9 M/s · 0.93 GiB/s | — |
| → pandas (Arrow) | 12.7 M/s · 1.09 GiB/s | — |
| → pandas (nullable) | 6.7 M/s · 0.58 GiB/s | — |
| → pandas (lazy/iter) | 10.8 M/s · 0.93 GiB/s | — |
| Arrow C-stream | 6.2 M/s · 0.53 GiB/s | — |

### Ingress — 10M rows
| role | py-pandas | rust-polars |
|---|---|---|
| CPU floor | 72.0 M/s · 6.69 GiB/s | 60.6 M/s · 5.63 GiB/s |
| e2e DataFrame→wire | 3.9 M/s · 0.36 GiB/s | 1.8 M/s · 0.17 GiB/s ⚠️ |

> **On comparing absolutes to the previous comment:** this box is running ~15–20% slower today than on Jun-22 — the pure-CPU **decode floors** dropped uniformly (py 17.5→13.3, rust 16.7→13.9 M/s) with no relevant code change on those paths. So the egress conversion absolutes here are all ~15–20% below the prior comment's; the regression *recovery* (4–7×) dwarfs that. Within-run COV is 0.3–5%.

> **⚠️ NB (separate issue, ingress — not part of this egress work):** Rust ingress e2e came in at a stable 1.8 M/s (identical across two runs, COV <1.5%) vs ~3 M/s previously, while **py ingress is unaffected** (~3.9 M/s) — so it's rust-specific, not a server-wide effect. Bisected to a single commit — see the **Update** below.

### Update — ingress e2e drop attributed to `6b0eb88`

Bisected the Rust ingress regression flagged above, on the same box: alternating OLD/NEW runs against one shared server (to cancel thermal/ordering), with the ingress example harness byte-identical across the range so the delta isolates the `column_sender/` library. It's a code change, not machine state — the pre-range binary still reproduces ~3.5 M/s *today*:

```
86ce1bc  baseline             3.57 M/s
49841f0  pooled handle close  3.49
ef8e10a  ffi pool refactor    3.43
6b0eb88  do some fixes        1.81   ← drop (0.53×)
1d88fc8                       1.82
```

**Cause:** `6b0eb88` adds `array_data.validate_full()` per polars column per flush in `ingress/polars.rs` (`ffi_polars_to_arrow_rs`, on the flush path). That's arrow-rs's full O(rows × cols) value scan — UTF-8 validation of the VARCHAR column, dictionary-key bounds on all 5 SYMBOLs, plus offset/null-buffer checks — across all 10M × 15, ≈ +2.6 s per flush, which is the whole 3.5 → 1.8 M/s gap. The **encode floor is unchanged** (~52–58 M/s before and after), so the cost is in the polars→Arrow conversion on the flush path, not in encoding.

Not simply removable: the unsafe varlen-offset fast path in `column_sender/arrow_batch.rs` now relies on it as a precondition (*"caller must uphold validate_full"*), and there's a second `validate_full` in `column_sender/chunk.rs`. So the cheaper-precondition route is the fix — a structural-only `validate()`, validating just varlen offset-monotonicity, or trusting the polars export and keeping checked access on the hot path.

Same methodology as above: one shared QuestDB (`9.4.3-SNAPSHOT` @ `85aecd600b`, `http.receive.buffer.size=16M`), single-stream, strictly sequential — Rust ingress populates the DEDUP table (`count()==10M` gate), Rust egress + Python egress read it back, Python ingress last. `GiB/s = rows/s × canonical on-wire bytes/row` (ingress 99.6744, egress 92.4641); native `MiB/s` is not cross-client comparable on egress (py records decoded-Arrow nbytes, rust on-wire), so `rows/s` is the basis. Aggregated with `doc/bench_parity_aggregate.py --ingress-bpr 99.6744 --egress-bpr 92.4641`.

jovfer:
## QWP DataFrame S1/S2 benchmark refresh — vs ENT 3.3.4 OSS core

Fresh local run against the **OSS core that Enterprise 3.3.4 ships** (`questdb` submodule pinned by the `3.3.4` tag), with the DataFrame clients held at the prior run's commits so the only variable is the server.

- QuestDB: `9.4.4-SNAPSHOT`, commit `5b2efe5e58fbb77ef25f87a7aa604365c9c1eb55` (= ENT 3.3.4 OSS core, on `master`)
- C/Rust client: `214830a9ac749800b41501d2722358a634a003ab` (pinned)
- Python client: `7334503e84e2d149f9d6550dd023ef484d2edc1e` (pinned); vendored `c-questdb-client` = `214830a9`
- Server launch: `-ea` (QuestDB's production default); single loopback instance, one client at a time (no parallelism)
- Shape: `10,000,000` rows; Rust `10` iterations, Python `5`; `2` warmups; full mode
- Validation: every real-server path verified `count() == 10,000,000` (DEDUP, no inflation)

Rates use binary units: `MiB/s = rows/s * canonical on-wire bytes/row / 1024^2`; `GiB/s = MiB/s / 1024`.

| Scenario | Direction | Client | Path | Rows/s | MiB/s | GiB/s |
|---|---|---|---|---:|---:|---:|
| S1 | ingress | Python | CPU floor | 523.39M | 22461.56 | 21.935 |
| S1 | ingress | Rust | CPU floor | 469.17M | 20134.72 | 19.663 |
| S1 | ingress | Python | e2e DataFrame->wire | 17.77M | 762.76 | 0.745 |
| S1 | ingress | Rust | e2e DataFrame->wire | 23.18M | 994.62 | 0.971 |
| S1 | egress | Python | decode floor | 64.82M | 2295.11 | 2.241 |
| S1 | egress | Rust | decode floor | 53.55M | 1895.98 | 1.852 |
| S1 | egress | Python | -> Polars | 49.15M | 1740.46 | 1.700 |
| S1 | egress | Rust | -> Polars | 52.37M | 1854.31 | 1.811 |
| S1 | egress | Python | -> pandas (default numpy) | 24.50M | 867.49 | 0.847 |
| S1 | egress | Python | -> pandas (Arrow backend) | 62.90M | 2227.16 | 2.175 |
| S2 | ingress | Python | CPU floor | 101.86M | 9682.33 | 9.455 |
| S2 | ingress | Rust | CPU floor | 64.18M | 6100.09 | 5.957 |
| S2 | ingress | Python | e2e DataFrame->wire | 5.42M | 515.08 | 0.503 |
| S2 | ingress | Rust | e2e DataFrame->wire | 7.25M | 689.12 | 0.673 |
| S2 | egress | Python | decode floor | 15.94M | 1405.54 | 1.373 |
| S2 | egress | Rust | decode floor | 12.32M | 1086.06 | 1.061 |
| S2 | egress | Python | -> Polars | 13.52M | 1192.24 | 1.164 |
| S2 | egress | Rust | -> Polars | 11.93M | 1051.65 | 1.027 |
| S2 | egress | Python | -> pandas (default numpy) | 14.44M | 1273.42 | 1.244 |
| S2 | egress | Python | -> pandas (Arrow backend) | 17.65M | 1556.18 | 1.520 |

Denominator note: the throughput columns use canonical QWP on-wire bytes/row per schema (S1 45.0 / 37.13 ingress/egress; S2 ~99.7 / ~92.5), not each client's raw egress byte counter — Python egress records decoded-Arrow bytes while Rust records on-wire bytes, so only `rows/s` (and the canonical-derived MiB/s) are cross-client comparable.

Main readout:

- **The server is not a regression.** A controlled comparison (same client binary, `-ea` on both, idle box, min-of-N) shows ENT 3.3.4's OSS core (`5b2efe5e`) ingests at the same speed as the prior run's server — S1 ingress min flush times were within ~1.3% across the June and July server commits. The earlier "server got 2x slower" reading was measurement noise, not code.
- Rust leads e2e ingress on both schemas: S1 `23.18M` vs Python `17.77M`; S2 `7.25M` vs `5.42M`.
- Pandas egress still hinges on the output backend: the default `to_pandas()` (numpy) is the slow path — S1 `24.50M`, ~2.6x behind the Arrow backend (`62.90M`) and behind `to_polars()` (`49.15M`). Selecting `dtype_backend="pyarrow"` recovers almost all of it.
- Egress CPU floors are close to the prior run on the Python side (client-bound, so server-insensitive). Rust's decode-floor came in below Python's this run (S1 `53.55M` vs `64.82M`); Rust and Python egress populate their own tables with different physical layouts, so this reflects table state / measurement rather than a client-capability change — flagged for follow-up.

Methodology caveat: single developer laptop, so ingest (server-write path) throughput is load-sensitive. Numbers were taken at low load (S1 ~load 2; the 15-column S2 self-heats the box to ~load 5, so S2 ingress is a floor, not a ceiling). Egress paths have tight cov (<0.03). Treat cross-run absolute deltas vs the prior comment as measurement/box-condition, not server behavior.

jovfer:
## QWP DataFrame benches over a real network — C vs Rust (AWS pair)

First numbers from the network rig (methodology + scripts: `doc/QWP_NETWORK_BENCH_PLAN.md`, `doc/net_bench/` on this branch). Loopback removed from the equation: dedicated server and client boxes, the DataFrame goes over a real link.

- Boxes: 2x `c8gn.2xlarge` (Graviton, 25 Gbps baseline), same cluster placement group, server data dir on tmpfs
- Channel (iperf3-gated before cells): single TCP flow **9.53 Gbps** (placement-group flow cap), RTT 0.11 ms, MTU 9001, 0 retransmits; cells below ran unshaped
- QuestDB: `9.4.4-SNAPSHOT`, commit `5b2efe5e` (= ENT 3.3.4 OSS core), `http.recv.buffer.size=16m`
- C/Rust client: both at `b4ca6575` (this branch); Rust = `qwp_ingress_polars` / `qwp_egress_polars`, C = the new `qwp_ingress_c` / `qwp_egress_c` twins (direct column sender, same schema/data/JSON contract)
- Shape: S2-wide, `10,000,000` rows; `5` iterations, `2` warmups; `MAX_BATCH_ROWS=10000`; egress populates then reads back (fresh reader per iteration, `compression=raw`)
- Validation: every real-server path verified `count() == 10,000,000` (DEDUP, no inflation); CoV <= 2.2% except C decode floor 4.7%

Rates use canonical on-wire bytes/row (S2: ~99.7 ingress / ~92.5 egress), same denominator note as the previous comments.

| Scenario | Direction | Client | Path | Rows/s | MiB/s | GiB/s |
|---|---|---|---|---:|---:|---:|
| S2 | ingress | Rust | CPU floor (encode) | 25.25M | 2400.7 | 2.344 |
| S2 | ingress | C | CPU floor (chunk build) | 12.19M | 1159.4 | 1.132 |
| S2 | ingress | Rust | e2e DataFrame->wire | 3.61M | 343.3 | 0.335 |
| S2 | ingress | C | e2e chunks->wire | 1.94M | 184.7 | 0.180 |
| S2 | egress | Rust | decode floor | 11.04M | 974.0 | 0.951 |
| S2 | egress | C | decode floor | 11.05M | 975.0 | 0.952 |
| S2 | egress | Rust | -> Polars | 9.77M | 862.3 | 0.842 |
| S2 | egress | C | -> materialized cells | 11.02M | 972.2 | 0.949 |

Main readout:

- **Egress is near wire-bound and the C client is at parity or better.** Decode floors are identical (both clients sit on the same decoder core), and C's materialize path adds ~nothing on top, landing ~13% above Rust's `-> Polars` (which pays DataFrame assembly). At ~0.95 GiB/s both are pushing ~85-90% of the measured 9.53 Gbps single-flow cap — the "can clients even use the pipe" question is answered for egress.
- **Ingress: Rust 3.61M vs C 1.94M rows/s — and the gap is client CPU, not the client's wire path.** Per pass the C client burns 4.64 s CPU out of 5.15 s wall (90%, i.e. throughput ~= its own CPU ceiling) vs Rust's 1.46 s of 2.77 s (53%, pipeline-bound). Root cause found: the raw chunk API re-resolves the full symbol dictionary per chunk (O(dict x batches) + a hash intern per referenced entry), while the Arrow/polars path reuses a buffer-identity memo (`take_arrow_dict_memo`) that the raw path can't reach. Two isolation probes confirm it: dict cardinality 100k->1k (same binary/rows/batches) takes C to **5.47M rows/s — above Rust**; batch 10k->100k only trims the scan term (2.39M). Detailed writeup + API proposal coming as a separate issue.
- Rust at this commit matches its earlier rig numbers at `214830a9` (3.61M vs 3.46M ingress, 9.77M vs 9.92M `-> Polars`) — no client regression from the `qwpws::`->`ws::` conf change or the connect-event work. One FYI: Rust's ingress encode floor dropped 34M->25.2M rows/s between those commits on the same box; off the e2e critical path and the example was rewritten in between, so possibly a measurement change — flagged, not investigated.
- Not run over the network yet: S1 (parity already shown on loopback), rate/RTT sweeps (P2/P3 of the plan), Python cells (still blocked on the harness `QDB_HOST` patch).

bluestreak01 (approved):

PR #156, opened 2026-06-10
chore: drop QWP per-table schema id, always inline column schemas

mtopolnik:
## Summary

Mirrors the server-side (questdb/questdb#7200) and Java-client QWP simplification in the Rust client: the wire format drops the per-table-block schema mode byte and schema id, so column schemas always travel inline.

### Ingress
- `QwpBuffer` writes columns inline right after `col_count` (no mode byte, no schema id) on every encode path — delta-dict, replay, and single-table datagram encoders.
- Drops the test-only `SchemaRegistry`, the `intern_replay_schema_signature` replay-id allocator, and the `QWP_SCHEMA_MODE_*` / `QWP_INLINE_SCHEMA_ID` constants.
- QWP/WebSocket connect strings now reject unknown keys per the connect-string spec (matching the .NET and C/C++ clients), tolerating the egress query-client keys so one `ws::` string drives both the sender and the `QwpQueryClient`. Legacy ILP transports keep ignoring unknown keys.
- `max_schemas_per_connection` was never part of a released client, so it falls through to the generic unknown-config-key path rather than getting a dedicated rejection arm.

### Egress
- The schema rides the first `RESULT_BATCH` (`batch_seq == 0`) of a query, parsed by the new `Schema::decode_inline`; continuation batches carry rows only and reuse it.
- The connection-scoped `SchemaRegistry` is replaced by a per-query `Option<Schema>` on the `Reader`, cleared at query start and on failover.
- `decode_result_batch` / `decode_frame` lose the registry parameter, `DecodedBatch` loses `schema_id`, and `CACHE_RESET` keeps only `RESET_MASK_DICT`.
- The frame header now validates `version == 1`. Versions collapse to a single value: `HIGHEST_KNOWN_VERSION` 2 → 1, and the egress handshake advertises `X-QWP-Max-Version: 1`.

### Removed and changed public API

None of this surface has shipped in a published crate (the `egress` module postdates the last release), so there is no semver impact — listed for review visibility:

- `Reader::schema_registry()` — removed (the registry no longer exists).
- `egress::wire::RESET_MASK_SCHEMAS` re-export and `resets_schemas()` — removed; bit `0x02` of `CACHE_RESET` is reserved and ignored.
- `DecodedBatch.schema_id` public field — removed.
- `#[doc(hidden)] _bench_internals` re-exports `Schema` instead of `SchemaRegistry`; `decode_result_batch`'s registry parameter becomes `&mut Option<Schema>`.
- `max_version` connect-string values collapse to `1`; a conf string with `max_version=2` (previously the default) now fails with `ConfigError` on the reader.
- `max_schemas_per_connection` loses its dedicated sender error: now the generic unknown-key rejection on QWP/WS strings, silently ignored on legacy ILP strings, and rejected as unknown by the reader (dropped from its ingress-key tolerate list).
- New public items: `egress::wire::header::PROTOCOL_VERSION` and `Schema::decode_inline`.

### Tests
- Refreshed byte-level fixtures (golden Java replay payloads, the shared datagram decoder, the WS replay schema assertions) for the inline format.
- Replaced the schema-reference decoder test with a continuation-reuse test plus a continuation-before-schema rejection test.
- Added a QWP/WebSocket unknown-key rejection test.
- Deleted the obsolete schema-id-exhaustion and reference-mismatch probes.
- Continuation-path coverage: both fuzz harnesses now sweep rows-only continuation frames against a retained schema; the fragmentation suite streams a batch-0 + continuations split; the failover mock gained `Action::SendBatch`, pinning post-failover schema re-read, continuation-before-schema rejection on replay, and the data-delivered duplicate guard; the C++ suite asserts continuation row values through the C API; `FrameHeader::parse` version rejection and the handshake `max_version + 1` boundary are unit-tested.
- C-header/FFI doc-contract fixes ride along: `line_reader_cursor_add_credit` added to the batch-invalidation lists, and the SERVER_INFO availability docs updated to the always-sent semantics.

## Summary by CodeRabbit

* **Refactor**
  * Protocol pinned to version 1; schema-mode/id removed and inline column definitions are emitted only in the first batch of a query. Reader now tracks schema per-query instead of a shared registry.

* **Bug Fixes**
  * Continuation batches are rejected if schema hasn’t been initialized; cache-reset now only clears dictionaries.

* **Configuration**
  * Connect-string parsing tightened; unknown QWP/WebSocket keys are rejected and default max-version set to 1.

* **Tests**
  * Many protocol, mock, fuzz, and golden tests updated to the new wire format and behaviors.


mtopolnik:
# Review of PR #156 (level 3)

**Verdict: approve.** All 10 review agents came back with **zero critical findings and zero broken callsites**. The full verification matrix ran green: 1,566 lib tests, 67 failover tests, TLS + both fuzz suites, decoder bench, FFI crate build, clippy with the new CI feature combo (zero warnings), `cargo fmt --check` clean, and all four C++ binaries (194 cases) against a rebuilt staticlib. Wire-format parity was independently confirmed against the pinned questdb submodule in both directions, and the hand-edited Java golden hex was decoded byte-by-byte and is exact. The asks below are test-coverage and doc-contract items — recommended before merge but none is a functional defect.

## Critical

None.

## Moderate

**M1. `line_reader.h` batch-invalidation contract omits `line_reader_cursor_add_credit` — documented-UB trap for C callers.** *(out-of-diff, pre-existing text, hazard live)*
`include/questdb/egress/line_reader.h:1108-1111` and `:1121-1124` list only `next_batch`/`cancel`/`free`/failover as invalidating a batch handle, but `line_reader_cursor_add_credit` routes through `cursor_for_mut()` (`questdb-rs-ffi/src/egress.rs:2447-2451`, call at `:2914`), which unconditionally drops the live `BatchView` — and its Rust core can trigger failover, clearing the `query_schema` slot the batch's column-name pointers borrow from. The C++ header gets it right (`line_reader.hpp:2210-2211`, `:2430-2432`), so the two headers contradict each other. A C caller following the `.h` text can hold column-name pointers across `add_credit` → use-after-free. Fix: add `add_credit` to both `.h` lists and the FFI doc at `egress.rs:2500`/`2960`.

**M2. SERVER_INFO availability docs now state the inverse of the implementation.** *(introduced by this PR's semantic flip; ~13 sites, mostly out-of-diff)*
The PR makes SERVER_INFO always-sent (`reader.rs:334`; the `Reader::server_info` doc was correctly updated), but every other C-facing doc still says it's absent on v1: `line_reader.h:430, 558, 715, 1383`; `line_reader.hpp:355, 580, 768, 2503`; `questdb-rs-ffi/src/egress.rs:835, 1262, 1577, 2716`; plus `reader.rs:823-824` (`FailoverEvent::new_server_info` "only on QWP v2+ servers" — read literally, never `Some`) and `reader.rs:1687-1689` (`Cursor::server_info` "None on v1 servers"). A C caller branching `if (!info) /* v1 */` per these docs now writes a branch that means "reconnect in flight". Reword to "NULL only during reconnect / if the server omits SERVER_INFO".

**M3. New continuation-batch state machine and version gate are under-tested.** *(in-diff)*
The decoder unit tests added (continuation reuse, continuation-before-schema) are good but the surrounding regression net has holes:
- `FrameHeader::parse`'s version≠1 rejection (`wire/header.rs:89-97`) — the PR's frame-level safety net — has **zero test coverage**; every helper that used to send version 2 was flipped to 1, so the rejected value is never exercised. One unit test (byte[4]=0 and =2 → ProtocolError) closes it; optionally pin `HandshakeVersion(2)` → HandshakeError in `egress_failover.rs` (the existing test uses 99).
- Both fuzz harnesses never reach the rows-only continuation path: `qwp_egress_bounds_fuzz.rs:368` hardcodes `batch_seq=0` and recreates a fresh `Option<Schema>` per decode, so the brand-new decode state gets no truncation/corruption sweep (Agent 2 probed it manually — no panics — but the harness wouldn't catch one introduced later).
- No test anywhere delivers a RESULT_BATCH **after** a failover reconnect: the `reader.rs:536` schema-clear is unobservable by the suite (delete it and everything still passes); the one data-then-failover C++ test (`test_line_reader_mock.cpp:3385`) has server B reply with only RESULT_END. Add a mock script where server B sends batch 0 (ideally a different shape) plus a negative variant whose first frame is seq=1.
- C-API continuation coverage is incidental (the seq 0..31 loop at `test_line_reader_mock.cpp:3138` asserts stats, not row values).

**M4. Fragmentation-fuzz helper builds malformed frames for any `batch_seq > 0`.** *(in-diff)*
`qwp_egress_fragmentation_fuzz.rs:125-149` (`result_batch_frame_seq`) writes `col_count` + inline schema unconditionally; under the new format that's only valid on batch 0. Saved today by its single caller passing seq 0 (`:322`) — but the parameterised signature advertises multi-batch support that now produces schema bytes where the decoder expects row data. The C++ twin was correctly gated on `batch_seq == 0`; mirror that (which also gives M3 its multi-batch fragmentation scenario for free).

**M5. Version-gate hardening: `X-QWP-Version: 0` (and legacy "1") pass the handshake and die slowly.** *(in-diff adjacent; `transport.rs:242` untouched)*
The handshake rejects only `server_version > max_version`. A server advertising "0" connects under default `target=Any`, skips the SERVER_INFO read (`reader.rs:334` `>= 1`), then every frame fails as a **failover-eligible** ProtocolError — the first query burns the entire failover budget re-dialing the same broken endpoint. A pre-#7200 server that negotiates old-semantics "1" (no SERVER_INFO frame) instead stalls `server_info_timeout_ms` (5 s default) per endpoint before erroring with a timeout rather than anything version-shaped — unavoidable at handshake since the version number was reused, but worth a doc note. One-line fix for the "0" case: reject `server_version != PROTOCOL_VERSION` at `transport.rs:242`, making the `reader.rs:334` conditional genuinely dead.

**M6. Shared-connect-string contract: `auth` is allowlisted but silently dropped by the sender, and the allowlist has no cross-check.** *(in-diff)*
`auth` is consumed by the reader as a verbatim Authorization header (`egress/config.rs:822`) but the sender has no `auth` arm — a shared `ws::` string whose only credential is `auth=...` produces an unauthenticated sender that fails with a confusing 401 at upgrade rather than a config error. Pre-existing silent-ignore, but the PR codifies it in `QWP_WS_PORTABLE_CONFIG_KEYS` without resolving the asymmetry. Also: nothing (test or build check) ties the hand-maintained portable list (`ingress.rs:746-774`) to the keys `ReaderConfig::from_conf` actually accepts — the in-code "MUST be reflected here" comment is unenforced; the new rejection test covers only 5 of 26 keys. Suggested: loop the egress key list over a `qwpws::` prefix in a test. Finally, the `from_conf` rustdoc (`ingress.rs:631-668`) never mentions that unknown keys are now rejected on `ws::`/`wss::` while remaining ignored elsewhere — that's the user-facing contract and only code comments document it.

**M7. PR description omits the removed public API.** *(process)*
The description covers the wire format well but never mentions: `Reader::schema_registry()` removed (the only removal on a flagship public type), `egress::wire::RESET_MASK_SCHEMAS`/`resets_schemas()` removed, `DecodedBatch.schema_id` removed, `max_version=2` now a hard ConfigError on the reader, and `max_schemas_per_connection` on **legacy** strings flipping from a descriptive error to silently ignored. All unreleased surface (crate 7.0.0 unpublished; latest tag 6.1.0 predates `egress`), so no semver violation — but add a "Removed public API" bullet and consider linking questdb#7200. The `chore:` prefix undersells a wire-format change (nit).

## Minor

1. **`reader.rs:1435-1439`** — `.expect("schema populated by inner decode")` contradicts the inner loop's own rationale at `:1579-1588` (which deliberately uses a soft ProtocolError "so an internal-invariant violation can't abort the process across the FFI boundary"; FFI is `panic=abort`). Provably unreachable today; make the wrapper mirror the inner's pattern so a future refactor can't re-open the abort path.
2. **`server_event.rs:368`** — `#[cfg(test)]` helper still constructs `FrameHeader { version: 2 }`; inert (tests bypass `parse`), but it would trip `write`'s debug_assert if ever serialized. Use `PROTOCOL_VERSION`.
3. **Doc rot sweep** (all verified): `reader.rs:404, 1389, 1511, 1601, 1826, 2578-2580`; `egress/mod.rs:31`; `server_event.rs:119, 143`; `decoder.rs:42-47` (module-level wire sketch still shows unconditional col_count+schema); `error.rs:129, 202, 241, 273`; `config.rs:396`; `msg_kind.rs:55`; `capabilities.rs:27-29`; `egress_failover.rs:122-124` (**factually wrong**: "0x02 clears the schema registry") and `:775`; `egress_live_server.rs:2798`; `egress_live_server_fuzz.rs:1205`; `ingress/tests.rs:306` (test string still uses now-invalid `max_version=2`) and `:322-326` (comment claims on_* keys ride the now-removed fallthrough); `line_reader.h:156` + ffi `egress.rs:87` ("schema/symbol-dict **reference** miss").
4. **`config.rs:78` / `wire/header.rs:43`** — `HIGHEST_KNOWN_VERSION` and `PROTOCOL_VERSION` are two unlinked constants both pinned to 1; bumping one without the other makes the handshake advertise a version the frame parser rejects. Define one in terms of the other. `PROTOCOL_VERSION` is also missing from the `wire/mod.rs:40` re-exports.
5. **`schema.rs:98`** — `Schema::decode_inline` is `pub` but has exactly one production caller (`decoder.rs:448`); `pub(crate)` suffices.
6. **`qwp_ws.rs:1087-1090`** — latent off-by-one: asserts `encoded_len >= 1024` then `max_buf_size(encoded_len - 1)`, whose hard minimum is 1024 (`ingress.rs:1900-1906`); at exactly 1024 the builder `.unwrap()` panics before the intended rejection. Guard should be `> 1024`.
7. **`egress_live_server.rs:~3384`** (Path B) — asserts post-cancel `Ok(None)`, which holds only while the server completes `select 1` before observing CANCEL; if cancel ever wins, the drain stashes `Err(Cancelled)` and the assertion fails. Near-deterministic today, but racy-by-construction in a suite this PR just promoted into CI.
8. Test naming: 89 occurrences of `let mut reg: Option<Schema> = None` — "reg" no longer refers to anything; mechanical rename to `schema` while the files are hot.
9. `cpp_test/qwp_mock_server.hpp:150,159` — dead `schema_id` parameter kept for call-site compatibility; fine, but consider dropping in a follow-up. `TestQwpUdpSender` isn't version-gated for `--repo` runs against pre-#7200 checkouts (acceptable for a dev flag; a comment would do). Java golden hex is verified-correct but now hand-derived — regenerate from the Java client once its inline-schema encoder lands so parity is proven, not asserted.
10. Commit nits: `d432dce9` is 52 chars and the only conventional-commit prefix on the branch; "De-rot" (`92d6fbff`) is project slang.

## Downgraded (false positives, dropped after verification)

- **Stale-frame schema corruption** (a prior query's `batch_seq=0` overwriting `query_schema` mid-new-query): unreachable as silent corruption — the rid check at `reader.rs:1544` fires immediately after decode and `terminate_with_close()` tears down the transport; every path that could leave stale frames readable also poisons/closes the transport; `execute()` re-clears the slot. Verified independently by four agents and me.
- **FFI dangling schema pointer via cancel/add_credit drain**: every mutating FFI op routes through `cursor_for_mut()`, which drops the `BatchView` before any decode can overwrite the slot — same discipline that previously protected against registry HashMap rehash. Only the `.h` doc gap (M1) remains.
- **Ingress size-estimate desync**: all estimator/encoder pairs dropped exactly the same 2 bytes; WS `payload_len` headers are patched from actual bytes written; estimate==encoded equality is proptest-enforced (`qwp.rs:8898-8922`).
- **OOM-abort via hostile `col_count`**: capped at 2048 before `decode_inline`; worst-case retention ~380 KiB.
- **Production wire-size regression from losing reference mode**: reference mode was `#[cfg(test)]`-only on main; both production encoders always emitted full schemas — production frames *shrink* by 2 bytes and the replay encoder loses a per-block linear scan.
- **CACHE_RESET schemas-bit loss**: clearing `query_schema` is structurally impossible now; ignoring bit 0x02 is exactly the spec's reserved-bit rule.
- **i64::MIN sentinel flip papering over a client bug**: no — densify-to-sentinel pre-exists on main with a unit test asserting `i64::MIN`; the live test had rotted while out of CI.
- **`ServerSchemaMismatch` error code dead**: still live (wire QUERY_ERROR status 0x03); never registry-related.
- **Bench semantics drift**: old and new benches both pay a full schema parse per iteration; numbers comparable.
- **`framed(2, ...)` stragglers / cpp mock seq>0 callers**: repo-wide greps empty; the only seq>0 caller (stats loop, seqs 0..31) is exactly the shape the new decoder expects.

## Summary

- **Verdict: approve** — the strongly-recommended pre-merge items are the cheap M3 coverage tests (version-rejection unit test, fuzz continuation coverage, post-failover batch test) and the M1/M2 C-header doc fixes, since those are the ABI contract text.
- No regressions found in the wire format, decode state machine, FFI safety, concurrency, resources, or performance; the one-slot-per-query design is strictly safer and faster than the registry it replaces. The deliberate tradeoffs (per-query inline schema bytes, unknown-key rejection on `ws::`, `max_version=2` now invalid) match the server/Java/.NET direction and are acceptable for unreleased surface — but deserve the M7 disclosure bullet.

mtopolnik:
PR review addressed as follows:

- fixed 5 of 7 moderate issues: 
  - skipped M5, deemed minor (defense against server announcing QWP version 0)
  - deferred M6 (`auth=` config keyword) to a separate cross-client effort to clean up both the spec and the implementations. We should remove `auth=` as a supported key.
- will add more commits for minor issues

mtopolnik:
# PR Review, Round 2

**Level 3** (all 10 agents, per-finding verification). **Verdict: approve with minor follow-ups.**

## Critical
None.

## Moderate

### M1 — `ws::` sender rejects ILP-transport keys *only when the matching transport feature is off* (in-diff)
`questdb-rs/src/ingress.rs:996-999` (new this PR) rejects any QWP/WS key not matched by an arm and not in `QWP_WS_PORTABLE_CONFIG_KEYS`. The arms for `request_timeout`, `retry_timeout`, `request_min_throughput`, `retry_max_backoff_millis` are `#[cfg(feature = "sync-sender-http")]` (`ingress.rs:972-989`); the UDP keys `max_datagram_size`/`multicast_ttl` are `#[cfg(feature = "_sender-qwp-udp")]` (`ingress.rs:784-787`). None are in the portable list.

Consequence: the *same* `ws::`/`wss::` connect string carrying e.g. `request_timeout=5000` **parses Ok with `sync-sender-http` compiled in** (hits the arm, value applied/ignored downstream) but **fails with `Unknown config key "request_timeout"` without it**. The default/CI feature set (`almost-all-features`) has http on, so CI never sees the rejection — it surfaces only in a qwp-ws-only build. The egress reader tolerates all of these unconditionally via `INGRESS_ONLY_CONFIG_KEYS` (`egress/config.rs:531-534, 518-519`), so the portability contract is asymmetric: reader accepts, feature-trimmed ws:: sender rejects.

Suggested fix: add the ILP-HTTP/UDP tuning keys to `QWP_WS_PORTABLE_CONFIG_KEYS` so a `ws::` string parses identically regardless of which sender transports are compiled in (mirrors how the reader already handles them). Empirically confirmed by Agent 9 and Agent 10 (both built `--features sync-sender-qwp-ws` without http and reproduced the rejection).

### M2 — `max_version` validity is asymmetric between reader and sender (in-diff, needs-decision)
The reader hard-rejects `max_version=2` (`egress/config.rs:771`, range `1..=HIGHEST_KNOWN_VERSION` = `1..=1`); the PR body documents this intentionally. But `max_version` has **no sender match arm** — it lives only in `QWP_WS_PORTABLE_CONFIG_KEYS` (`ingress.rs:764`), so the sender silently ignores it. A shared string `ws::addr=h:9000;max_version=2` therefore yields `SenderBuilder::from_conf` → **Ok**, `Reader::from_conf` → **Err**. Since `max_version=2` was the *default* before this PR, a previously-working explicit string now splits.

This is consistent with general portable-key semantics (the sender ignores all reader-only keys without range-checking them, e.g. `compression_level` too), and the reader-rejects-2 half is deliberate. So it is defensible as-is — but flagging because the PR explicitly markets "one `ws::` string drives both." Decide whether the sender should also range-validate `max_version` on QWP/WS strings for symmetry, or accept the documented asymmetry. (Verified by Agent 10 empirically.)

### M3 — No deterministic test for the stale-schema-across-query-boundary property (test gap)
The subtle risk of the single per-query `Option<Schema>` slot — query A drains, query B on the same connection must decode against B's schema, not A's — is **verified safe** in source (cleared at `reader.rs:1112` `ReaderQuery::execute`, plus `decode_result_batch` unconditionally overwrites on `batch_seq==0` and errors on continuation-before-schema), and is covered behaviorally by `egress_live_server.rs:1110 inline_schema_reread_per_query_on_shared_connection`. But that test needs a live Maven-built server; there is **no deterministic C++-mock or Rust-unit equivalent** issuing two sequential same-cursor queries with different column shapes. The C++ mock has all the primitives. Cheap insurance for a one-line invariant. (Agent 7.)

## Minor

### m1 — Portability test doesn't exercise the path it guards (test gap + latent dead code)
`ingress/tests.rs:271 ingress_silently_accepts_every_egress_only_key` is `#[cfg(feature = "sync-sender-http")]` and uses the `http::` scheme, so it passes through the always-accept `_ => builder` arm and never reaches the new QWP/WS strict-key branch. It therefore does not verify that `QWP_WS_PORTABLE_CONFIG_KEYS` actually covers every egress key, and `EGRESS_ONLY_CONFIG_KEYS` becomes an unused-const (clippy `never used`) in a `sync-sender-qwp-ws`-only build (CI's `almost-all-features` masks this). Fix: add a `#[cfg(feature = "sync-sender-qwp-ws")]` twin looping `EGRESS_ONLY_CONFIG_KEYS` over a `ws::` string. (The hand-checked list is currently complete both directions — this is about preventing future silent drift.) (Agents 9, 10.)

### m2 — Commit title lengths (informational)
8 of the branch's commit titles exceed the 50-char convention (e.g. `egress: prove Java golden parity against the inline-schema encoder` = 66). All are clear, plain-English, active voice — length only. (Agent 8.)

## Downgraded (false positives / verified non-issues)

- **"Ingress bandwidth regression from always-inline schema"** — the only encoder that ever used reference-mode (`QwpBuffer::encode_ws_message`) is **test-only** (sole caller `qwp_ws_protocol_probe.rs:395`). Both production paths (WS replay `encode_ws_replay_message`, UDP `encode_row_group_from_scratch`) already emitted full inline schema on `main`; this PR makes them **2 bytes smaller** per table block (drops mode byte + `schema_id` varint). Not a regression. (Agent 6, verified against `main`.)
- **Reachable panic in the decode path (process-abort under `panic="abort"`)** — none. `Schema::decode_inline` indexing is fully guarded (`checked_add`, `name_end > bytes.len()`, `cursor >= bytes.len()`); `Vec::with_capacity` is double-bounded (`col_count ≤ MAX_COLUMNS_PER_TABLE` at the caller + `.min(bytes.len())`); the two new `.as_ref().unwrap()` in `reader.rs:1453/1455` sit behind an `is_none()` early-return; `FrameHeader::write`'s `debug_assert` is library-internal and unreachable from any non-1 construction. (Agents 1, 2.)
- **`DecodedBatch.schema_id` removal breaks FFI/ABI** — no. Zero references to `schema_id` anywhere in `questdb-rs-ffi/` or `include/`; it was never exposed across `extern "C"`. The FFI/header diff is comment-only (no signature/discriminant/layout change). (Agents 3, 8.)
- **Stale-schema data corruption across query/failover boundary** — verified safe on all three start/replay paths plus the unconditional decoder overwrite + continuation-before-schema guard. (Agents 4/5, 9.)
- **Concurrency race on the single `query_schema` slot** — one-cursor-at-a-time is enforced statically in Rust (`&mut Reader`) and via the FFI `AtomicBool`; `Schema` is `Send+Sync`; removed ingress scratch state was test-only. (Agent 4/5.)
- **`add_credit` dangling-pointer doc** — accurate: `cursor_for_mut` (`egress.rs:2448-2452`) clears `current_batch` before mutation, and `Cursor::add_credit` can route through `failover_reconnect_and_replay`. The PR *closes* a pre-existing latent hazard (prior docs omitted `add_credit` from the invalidation list). (Agent 3.)

## Summary
- **One-line verdict: approve** — merge-ready; M1/M3/m1 are worth a quick follow-up commit, M2 is a design call.
- **Regression/tradeoff note:** the only behavior change reaching users is stricter QWP/WS connect-string validation (intentional, spec-aligned). M1 is the one place that strictness is applied *inconsistently* (feature-flag-dependent); everything else is internal refactor with byte-for-byte golden-test parity against the Java inline-schema encoder.

bluestreak01 (approved):

PR #158, opened 2026-06-17
test(qwp): assert empty vs NULL array distinction in the egress decoder, clean up stale comments

mtopolnik:
Does not contribute any production code, only a regression test and comment update.

Comments stated a divergence from Java implementation, because Java was violating the spec. Now Java has a fix so the comment is stale.

The regression test asserts the case at hand, ensuring that an empty array is seen as distinct from null array, and successfully decoded in server response.

## Changes

- Add `decode_array_empty_vs_null_distinct`: a view-layer regression test over regular / empty / NULL rows — the empty row is non-null with shape `[0]` and 0 elements; the NULL row has no shape.
- Restate the `array_dim_zero_is_valid_empty_array` doc comment declaratively, dropping the stale note about the Java client diverging from the spec (the Java client now accepts empty arrays too).

## Summary by CodeRabbit

* **Tests**
  * Added a decoder test to confirm correct handling of DOUBLE_ARRAY rows, distinguishing non-NULL empty arrays from NULL arrays (including shape and decoded raw payload expectations).

* **Documentation**
  * Updated inline guidance for DOUBLE_ARRAY edge-case behavior and clarified decoder validation notes for size-mismatch scenarios.

* **Chores**
  * Added a new PR review skill document defining a structured, adversarial review workflow and output format.

### DIFF
--- a/.pi/skills/review-pr/SKILL.md
+++ b/.pi/skills/review-pr/SKILL.md
@@ -1,0 +1,420 @@
+---
+name: review-pr
+description: Review a GitHub pull request against QuestDB client library coding standards. Use when asked to review a PR, given a PR number or URL, optionally with a review level 0..3. Performs an adversarial, blocking, mission-critical code review.
+allowed-tools: bash read subagent
+metadata:
+  argument-hint: "[PR number or URL] [--level=0..3]"
+---
+
+# Review a QuestDB client library pull request
+
+Review the pull request identified by the arguments you were invoked with. When
+this skill is run as `/skill:review-pr <args>`, the `<args>` are appended below
+as a `User:` message — treat that text as `$ARGUMENTS` everywhere this document
+refers to it. If no PR target is present in the arguments, ask the user which PR
+to review before doing anything else.
+
+> Harness note: this skill runs inside **pi**. Where the review needs parallel
+> review agents, use the `subagent(...)` tool with fresh-context `reviewer`
+> agents (see "Spawning review agents in pi" below) — there is no Claude `Agent`
+> tool. Where it needs to search the repository, use `bash` (`rg`, `grep`,
+> `find`) plus `read`; pi has no separate `Grep`/`Glob` tools.
+
+## Review mindset
+
+You are a senior QuestDB engineer performing a blocking code review. The QuestDB client library is mission-critical software — bugs can cause data loss, silent data corruption, or crashes in customer applications across Rust, C, C++, Python, and downstream language bindings. There is zero tolerance for correctness issues, resource leaks, undefined behavior, or unsound FFI. Be critical, thorough, and opinionated. Your job is to catch problems before they ship, not to be nice.
+
+- **Assume nothing is correct until you've verified it.** Read surrounding code to understand context — don't just look at the diff in isolation.
+- **The diff is a hint, not the boundary of the review.** The highest-value bugs almost always live at callsites outside the diff that depend on contracts the diff quietly changed. Treat the diff as the entry point, not the scope.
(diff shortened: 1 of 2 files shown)


mtopolnik:
# Review: PR #158 

**Level 2.** This is a test-only diff: +53/−6 in `questdb-rs/src/egress/decoder.rs`, entirely inside `#[cfg(test)] mod tests`. One new test plus a doc-comment edit. No production code, no `unsafe`, no FFI, no public API change.

## Critical
None.

## Moderate
None.

## Minor

**1. The same doc-comment still carries a stale C++ cross-reference — in-diff block, retained lines (`decoder.rs:4485-4489`).**
The PR's stated purpose is "clean up stale comments," and it correctly removes the stale Java-divergence note. But the retained middle of that same comment is itself stale on three counts:

| Comment says | Actual (`cpp_test/test_line_reader_mock.cpp`) |
|---|---|
| line `:1091` | line **1109** |
| `mock: non-null empty-data array row exposes data == NULL` | `mock: non-null empty-data array row exposes data_offsets symmetry` |
| "non-null row with `data == nullptr` and `element_count == 0`" | asserts `data_offsets` symmetry (zero-length slice) + `element_count == 0`; line 1141 explicitly **discards** the data pointer via `(void)col.elements<double>(0, &cnt)` — it no longer pins `data == nullptr` |

This is pre-existing (the PR didn't touch these lines), so it's non-blocking — but it's squarely inside the comment the PR is curating, and fixing the line number / test name / `data == nullptr` phrasing here would actually complete the de-staling the PR set out to do.

**2. (trivial, optional) Inline column bytes vs existing helper — `decoder.rs:4503`.**
`vec![0x01u8, 0x04]` duplicates what `col_with_bitmap(&[0x04], …)` (`decoder.rs:1911`) already does. Not worth changing — the inline form with its explanatory comment is clear and other tests build columns manually too. Noted only for completeness.

## Summary
**Approve.** The regression test is correct, meaningful, and verified to pass; the byte layout and all 13 assertions trace exactly to the decoder + view semantics; fmt and clippy are clean. The comment de-staling that was done is accurate and improves the declarative quality of the doc.

The only thing worth touching before merge is **Minor #1** — while removing one stale reference, the PR leaves a wrong line number (1091→1109), a wrong test name, and an outdated `data == nullptr` claim in the very same comment. It's non-blocking documentation, but it's directly in scope of the PR's stated goal.

bluestreak01 (approved):
Approve.

Test-and-comment-only change to the QWP egress decoder; zero production code changed. Verified:

- New `decode_array_empty_vs_null_distinct` traced byte-for-byte against `decode_array`, `is_null_at`, and `DoubleArrayColumn`/`ArrayLayout`; all 13 assertions match the decode path. Test is non-trivial (misclassifying empty↔null fails it) and runs green. Full decoder module: 76 passed. `cargo fmt --check` and `cargo clippy --tests` clean.
- Comment de-staling is accurate: `array_dim_zero_is_valid_empty_array` now cites the correct C++ test `mock: non-null empty-data array row exposes data_offsets symmetry` (test_line_reader_mock.cpp:1109), drops the rotted line number, and replaces the `data == nullptr` claim with 'zero-length (empty) per-row data slice' — matching what the C++ test asserts. The zstd comment now references `written != frame content size` in `zstd_decompress_body` instead of a stale line number.

Non-blocking: the PR also bundles a 420-line unrelated `.pi/skills/review-pr/SKILL.md` that the description doesn't mention ('only a regression test and comment update'). Worth splitting into its own PR or disclosing in the description, but not a merge blocker.

PR #160, opened 2026-06-18
ci: run "Vs QuestDB master" against the nightly docker image

ideoma:
## What

Replaces the from-source build of QuestDB master in the **"Vs QuestDB master"** integration job with the prebuilt **`questdb/questdb:nightly`** docker image. The nightly image tracks master, so the "did QuestDB master break the client?" signal is preserved while the Maven + Rust from-source build is dropped from this job.

Both the integration matrix and the mid-query failover suite now run against the container (the scope agreed up front).

## How

**Harness (`system_test/`)**
- `fixture.py` — new `QuestDbDockerFixture`, a drop-in for `QuestDbFixture`:
  - config injected via `QDB_*` env vars (no conf file mount needed);
  - container ports published **1:1 on loopback**, so a client still connects to `127.0.0.1:<port>`;
  - server log read via `docker logs` (the image logs to stdout); data is ephemeral and removed with the container on `stop()`;
  - ILP TCP auth via a read-only single-file bind mount at `conf/auth.txt`;
  - `stop()` → `docker stop` (graceful SIGTERM); `kill()` → `docker kill` (SIGKILL);
  - `_check_http_up()` tolerates a connection reset, because docker publishes the host port *before* QuestDB is ready (the local-process fixture never saw this).
- `test_egress_failover.py` — new `DockerStandaloneInstance` with the same crash semantics.
- `test.py` / `test_egress_failover.py` — new `--docker IMAGE` flag.

**CI (`ci/run_tests_pipeline.yaml`)**
- The job pulls `questdb/questdb:nightly` and runs `test.py run --docker` + `test_egress_failover.py run --docker`. No Maven/Rust QuestDB build.
- Moved back to Microsoft-hosted `ubuntu-latest` (docker preinstalled). The disk pressure that originally pushed this job to `hetzner-incus` was the from-source build + on-host server data, both now gone.
- The from-source `clone_questdb.yaml` template is unchanged and still used by the mac/windows/linux fuzz jobs.

## Why `docker kill`, not `docker stop`, for failover

The failover test deliberately SIGKILLs the primary mid-stream to simulate a crash with **no graceful WebSocket Close** — that ungraceful drop is what makes the client fail over. `docker stop` sends a trappable SIGTERM (the JVM closes connections cleanly), which would *not* engage failover. So the crash maps to `docker kill` (SIGKILL); only the genuinely-graceful path uses `docker stop`.

## Local validation (against `questdb/questdb:nightly`)

- Integration matrix: `auth` on/off, build modes (API/CONF/ENV), TLS proxy — pass.
- Mid-query failover: `docker kill` of server #1 → `failover_resets=1`, 1,000,000 rows replayed, final endpoint = server #2 — pass.
- No leftover containers after runs.

## Tradeoffs / things to check before merge

- **Pool choice.** `ubuntu-latest` guarantees docker but the disk-fill that moved this job to `hetzner-incus` was attributed to the fuzz portion. The from-source build (the bigger consumer) is gone, so this should be fine, but if disk pressure recurs the fallback is a higher-headroom pool that also has docker.
- **Required-check name changed.** The job is renamed `TestVsQuestDBMaster` → `TestVsQuestDBNightly` ("Vs QuestDB nightly (docker)"). If a branch policy lists the old check name as required, update it.
- **Coverage delta: QWP UDP.** It is gated on `--repo` in `test.py`, so it is not exercised in docker mode. That coverage still runs in the from-source jobs; not ported here to avoid an unvalidated UDP-over-docker path.

## Summary by CodeRabbit

* **Tests**
  * Added Docker container support for running system tests against prebuilt Docker images
  * Improved test infrastructure to support multiple execution modes

* **Chores**
  * Updated CI pipeline to execute tests using nightly Docker images instead of building from source
  * Enhanced failover testing to support both local and containerized execution

### DIFF
--- a/ci/run_tests_pipeline.yaml
+++ b/ci/run_tests_pipeline.yaml
@@ -151,112 +151,71 @@ stages:
           - script: ctest --output-on-failure -E rust_tests
             workingDirectory: build_sanitize
             displayName: "ctest under ASan + UBSan"
-      # Runs the full system_test/test.py against a fresh build of
-      # QuestDB master. Moved off Microsoft-hosted ubuntu-latest because
-      # the QWP/WS fuzz portion of the run filled the agent disk ("No
-      # space left", SIGBUS); the hetzner-incus pool — already used by
-      # TestQwpWsFuzzLinux below and by the cron-time fuzz pipeline —
-      # has enough headroom. Pool image is built from
-      # questdb/azure-ci-runners' roles/incus/files/provision.sh, which
-      # provides openjdk-17-jdk, openjdk-25-jdk, maven, build-essential,
-      # zip, curl, gh, and rustup; only cmake and python3-numpy are
-      # apt-installed per job. imageName='hetzner-incus' tells
-      # compile.yaml to skip its pip step.
-      - job: TestVsQuestDBMaster
-        displayName: "Vs QuestDB master"
+      # Runs the integration matrix (system_test/test.py) and the
+      # mid-query failover suite (system_test/test_egress_failover.py)
+      # against the prebuilt QuestDB *nightly docker image* instead of a
+      # from-source build of master. The nightly image tracks master, so
+      # this keeps the "did QuestDB master break the client?" signal while
+      # dropping the Maven + Rust from-source build entirely. The server
+      # runs in a container (config injected via QDB_* env vars, ports
+      # published 1:1 on loopback); the failover crash is a `docker kill`
+      # (SIGKILL), the graceful path a `docker stop` (SIGTERM).
+      #
+      # Back on Microsoft-hosted ubuntu-latest (which ships docker
+      # pre-installed): the disk pressure that originally pushed this job
+      # to hetzner-incus came from the from-source build plus on-host
+      # server data, both now gone (the server lives in an ephemeral
+      # container, removed between configs). If fuzz-portion disk
+      # pressure ever recurs, move this job to a pool with more headroom
+      # that also has docker.
+      #
+      # Scope note: the nightly image is Linux/amd64 only, so the
+      # mac/windows fuzz legs stay from-source, as does any run that must
+      # test a specific commit not yet in nightly. QWP UDP is not
+      # exercised here (it is gated on `--repo` in test.py); that
+      # coverage still runs in the from-source jobs.
+      - job: TestVsQuestDBNightly
+        displayName: "Vs QuestDB nightly (docker)"
         pool:
-          name: hetzner-incus
+          vmImage: "ubuntu-latest"
         timeoutInMinutes: 90
         variables:
-          imageName: hetzner-incus
+          QUESTDB_IMAGE: "questdb/questdb:nightly"
         steps:
           - checkout: self
             fetchDepth: 1
             lfs: false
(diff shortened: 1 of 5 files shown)


ideoma:
### CI fix: failover exhaustion tests (`--network host`)

First CI run: integration passed, but the failover step failed on the two **exhaustion** tests — after `docker kill` of every endpoint, `next_batch()` returned `Ok` instead of erroring (`exit 12`). `test_mid_query_failover` (single kill → failover to the live server) passed.

**Cause.** The QWP egress WS reader is app-level flow-controlled (it pulls batches), so a *local process* never pre-sends the next batch — the original from-source tests pass. A **published port** routes through `docker-proxy`, a userland copy-loop that eagerly drains the server stream into its own buffer regardless of the reader's pacing; those buffered batches stay readable after the container dies, so the "every endpoint dead → next read must fail" contract breaks.

**Fix.** Run the failover containers with `--network host`. The server binds the discovered ports directly on the host stack and the client talks straight to it on loopback — no `docker-proxy` — i.e. the exact path of the local-process fixture these tests were written against. The integration fixture keeps published ports (it passed; its restart suite doesn't depend on a read failing the instant the server dies).

**Caveat (honesty):** Docker Desktop on macOS can't reach a host-networked container (VM boundary), so I could not reproduce this fix locally — it is validated only on the Linux CI agent. The CI re-run on this commit is the validation. (Consequence: docker-mode failover requires a Linux host; on macOS use `--repo`.)

ideoma:
### Correction + real fix for the exhaustion failures

My previous comment blamed `docker-proxy` buffering and switched the failover containers to `--network host`. **That was wrong** — host networking did not fix it (build 243434 failed identically). Apologies for the misdirection; here is the verified root cause.

**Root cause.** A diagnostic showed `next_batch()` after the kill returns `Ok(Some)` with a **full 16,384-row batch already in the client's buffer**. The QWP egress reader replenishes credit as it drains, so the server streams ahead. The optimised release **container** server (ParallelGC, warm) wins that prefetch race — landing the next batch in the client before `docker kill` lands — where the slower **from-source** server (G1, `-ea`) does not. So it's a latent timing assumption in the tests, exposed by the faster server; not a proxy/harness bug. (Recent `main` builds confirm the from-source job passes these same tests.)

**Fix.** In `exhaustion_client.rs`, after the kill, **drain the already-received batches, then require the reader to surface an `Err`** once it must touch the dead network. This still catches the real contract violations — `Ok` forever (capped), a clean `RESULT_END` terminal, or a panic on the poisoned reader — and the error code stays in the accepted set (`SocketError`/`ProtocolError`/`FailoverWouldDuplicate`). It also hardens the from-source run against the same race.

**Networking.** Since the helper no longer assumes zero buffering, `--network host` is unnecessary, so I reverted the failover containers to 1:1 published ports — same as the integration fixture, and it restores macOS support (Docker Desktop can't reach a host-networked container).

**Validated locally** against `questdb/questdb:nightly` on the heavier-buffering published-ports path: all four failover tests pass (`Ran 4 tests, OK`).

jerrinot (approved):
looks good, can merge on green.

PR #164, opened 2026-06-24
fix(system_test): use min-HTTP health check in fuzz test instead of /ping on main HTTP

mtopolnik:
## Problem

`TestQwpWsFuzz.test_all_mixed_with_bounce` flakes on CI with:

```
producer ... failed: QWP/WebSocket close drain timed out before all published frames were acknowledged
```

The fixture log made it look like a QuestDB bounce restart took ~113s (vs the usual 6–40s in the same run), which consumed the producers' 120s `close_drain` budget.

## Root cause

The server actually started in ~9s; the fixture's `/ping` readiness probe was then starved for ~100s.

`/ping` is served by the main HTTP server, which runs on QuestDB's shared worker pool — the same pool that accepts and processes QWP ingest. On a bounce restart the reconnecting fuzz producers immediately replay their queued backlog and saturate that pool on the 4-core CI box, so the accept loop never runs and no new connection (including `/ping`) is accepted for ~80s. The probe could only succeed once the producers gave up at their `close_drain` deadline and disconnected — the server only looked "up" again after the producers had already failed.

## Fix (original scope)

- **Each start phase gates readiness on the server it actually uses.** A bounce restart probes `/status` on the min-HTTP server (`http.min.enabled=true`, dedicated single-worker pool, published as a 4th fixture port), which reconnecting producers cannot starve. The initial cold start keeps probing main-HTTP `/ping`: all test SQL (starting with `query_version()`) runs on the main server, the min-HTTP pool comes up seconds earlier, and no ingest exists yet that could starve `/ping`. Both probes share one implementation on `QuestDbFixtureBase`, and the probe raises immediately if the server process has died instead of retrying into the timeout.
- **Timing budgets sized to a full bounce.** A bounce keeps the server unavailable for a whole `stop()` + `start()` cycle, so the producers' reconnect and `close_drain` budgets derive from `bounce_stop_timeout_s` (120s) + `BOUNCE_RESTART_TIMEOUT_SEC` (90s) + headroom = 270s. A slow-but-healthy bounce can no longer surface as a misleading client-side `close_drain` timeout, and a genuinely stuck restart fails fast as a server lifecycle error.

## Additional fixes picked up while stabilizing CI

Each CI failure along the way exposed a distinct pre-existing defect, mostly in test infrastructure. Their fixes are part of this PR.

### Windows hosted runners: Ctrl+C never reached the JVM

Every server-bounce test on hosted Windows force-killed the JVM after the full stop budget. The fuzz supervisor launches its isolated child with `CREATE_NEW_PROCESS_GROUP`, which sets the "ignore Ctrl+C" flag for the group — and the JVM inherited it, so the graceful-shutdown Ctrl+C that `stop()` posts was ignored.

- `fixture.py` clears the inherited flag (`SetConsoleCtrlHandler(None, False)`) before launching the JVM; graceful shutdown completes in ~3s again.
- The console-ctrl helper that posts the event now deterministically survives its own Ctrl+C (it sets ignore-Ctrl+C after `AttachConsole` and sleeps until delivery completes), removing the spurious "Failed to post ctrl_c" reports.
- On a stop timeout, `stop()` dumps the last 32KB of the server log to stderr (`print_log_tail`) — the diagnostic that identified the flag inheritance in the first place.

### Fuzz lifecycle ownership and bounded waits

- A failed bounce no longer tries to recover inside `BounceThread`. The thread records the error and exits; the test joins it and then recovers the fixture synchronously (`finish_bounce_thread`), so lifecycle work cannot race verification or teardown on shared fixture state.
- `stop()` bounds the post-force-kill reap; `QuestDbStopTimeout.fixture_reusable` / `QuestDbReapTimeout` encode whether a recovery `start()` is safe afterwards.
- The fuzz supervisor bounds every lifecycle wait and force-stops the whole isolated process tree on its deadline (`killpg` on POSIX, `taskkill /T /F` on Windows), so a stuck JVM cannot consume the CI job timeout.
- The Arrow SFA bounce tests follow the same contract: they restart the shared server whenever `stop()` reaped it, `ArrowFuzzBase.setUp` restarts a server that a previous test left down, and every explicit `column_sender_sync` has a finite no-progress deadline. One failed bounce now fails one test instead of hanging the suite to the 1800s supervisor deadline.
- The Arrow SFA bounce producer's reconnect/close-drain budget is sized to the same 270s stop+start window as the QWP/WS producers, and it no longer pairs `initial_connect_retry=sync` with `lazy_connect=true` (rejected at build time under the java-parity pool-startup rules).

### ctypes egress reader: Windows access violation

The FFI returns borrowed, non-NUL-terminated column-name/error strings with their length in an out-param. Reading them via `c_char_p.value` makes ctypes scan for a NUL terminator past the end of the slice, which crashed on Windows with 0xC0000005. The reader now binds the out-param as `POINTER(c_void_p)` and copies exactly `len` bytes; a new guard-page regression test crashes without the fix.

### Live-server `/exec` requests over a raw socket

http crate 1.5.0 (released mid-branch; CI resolves dependencies fresh) started enforcing its 64 KiB total-URI cap, which ureq converts into a panic. QuestDB's `/exec` endpoint reads SQL exclusively from the URL query string, so the egress fuzz suites' multi-hundred-KB `INSERT` URLs are permanently out of reach for any client built on the http crate. `http_exec` now writes the GET request over a raw `TcpStream`: SQL percent-encoded into the request line, response read until structurally complete (terminal 0-chunk or `Content-Length` bytes — the server keeps connections alive, so reading to EOF would hang), 30s read/write timeouts.

### Smaller CI fixes

- clippy 1.97's new lints failed the fmt+clippy job for every build: `byte_char_slices` in the sender tests and `useless_borrows_in_formatting` in `tls_proxy`. Both fixed.
- The fake-server shutdown unit test launches its child with `CREATE_NO_WINDOW`, so the Ctrl+C it exercises no longer reaches (and kills) the unittest runner on a real Windows agent.
- The `questdb` submodule moves forward 4 upstream commits (including the fix that stops a query when its HTTP client disconnects).

### Unit coverage and CI wiring

- +33 server-free unit tests (41 → 74 across the four `system_test/test_*_unit.py` files): readiness-probe selection and the 204-vs-2xx status predicates, stop/reap ownership, Windows Ctrl+C flag handling, log-tail dumping, bounce lifecycle ownership, supervisor tree-kill, the Arrow shared-server recovery helpers, and the new ctypes reader suite (including the guard-page test).
- `ci/run_all_tests.py` runs the fixture/fuzz suites in `unit` and `cpp` modes and the ctypes reader suite in `all`/`unit`/`cpp`, so the Windows-only code paths execute on real Windows agents. `doc/BUILD.md` documents the invocation.

## What this does not change

The readiness probe is now isolated from ingest starvation, but the producers still reconnect QWP on the shared pool, and a bounce's graceful stop still drains WAL — on a severely CPU-starved box both remain slow. The sized budgets keep the blame accurate: a slow-but-healthy bounce passes, and a genuinely stuck server fails as a server lifecycle error rather than a client timeout.

## Test plan

- All CI checks are green, including the QWP/WS fuzz suite on linux, mac, and hosted Windows (the jobs that originally flaked), ASan+UBSan, the QWP egress live-server suite, and Vs QuestDB master.
- `python3 -m unittest` on the four unit files: 74 tests pass (fixture 19, QWP/WS fuzz 23, arrow-common 29, ctypes reader 3).
- Live smoke tests against a local QuestDB build: min-HTTP `/status` answers 200 from its dedicated pool while main `/ping` answers 204; with `http` locked to 1.5.0, the ureq-based `http_exec` reproduces the CI panic and the raw-socket version passes all live-server suites (egress fuzz 4/4, fragmentation 4/4, bind 4/4, ALTER 1/1, functional 82/82).


mtopolnik:
**Reviewing at level 3.** Note up front: this PR is **entirely Python test infrastructure** (`system_test/fixture.py`, `qwp_ws_fuzz.py`, `test.py`). None of the level-3 machinery aimed at FFI/ABI, ILP wire format, `unsafe`, panics, or the C/C++ headers applies — there is no Rust/C/C++ in the diff. I ran the dimensions that *do* apply: correctness/timing math, process/file-handle lifecycle, cross-test blast radius, test coverage, and a fresh-context adversarial pass.

**The core fix is sound.** The root cause (main-HTTP `/ping` starved because it shares the `shared-network` worker pool with QWP ingest, so "server back up" was structurally gated on producers *failing*) is correctly addressed by moving the bounce-restart probe to a dedicated-pool min-HTTP `/status`. The budget arithmetic is internally consistent: `client_budget = wind_down = bounce_stop_timeout_s(120) + BOUNCE_RESTART_TIMEOUT_SEC(90) + 30 = 240s`, which clears the worst legitimate down-window (stop <120s + restart <90s ≈ <210s) by the +30s headroom, so a healthy-but-slow bounce can't strand a producer mid-drain and an over-cap leg is attributed in the bounce thread. Managed/docker config did not drift. `int()` correctly keeps `240000.0` out of the conf string.

## Critical
None.

## Moderate

**M1 — Zero unit-test coverage for the new readiness logic (in-diff).** `fixture.py:381-418` (`_check_main_http_up`, `_check_min_http_up`, `_probe_http`, the three `_assert_server_alive` overrides) and the `start(start_timeout_sec, probe_min_http)` selection wiring have **no** direct test. The PR's "19/19 pass" is true but every one of those 19 is a pre-existing test for `stop()` escalation / `print_log()` / `is_transient_network_error` — none touches the new code (verified by reading `test_fixture_unit.py` and grepping the symbols). The single most valuable gap is the **status-predicate asymmetry** — main requires exactly `status == 204`, min accepts `200 <= status < 300` (`fixture.py:388,398`). A future edit that flips these ships green and silently breaks readiness gating, which is the whole point of the PR. The existing harness makes this cheap: `test_fixture_unit.py:61-64,93` already builds a `QuestDbFixture` from a temp dir and assigns a stand-in `_proc`. A server-free test can monkeypatch `urllib.request.urlopen` to assert the True/OSError→False branches, assign a finished `_proc` to assert `_probe_http` fails fast via `_assert_server_alive` before any HTTP call, and call the two `_check_*` methods with `_probe_http` captured to pin `(port, path)` and `status_ok(204/200/500)`. Non-blocking (the fix is validated by the integration run + the author's live smoke test), but worth adding given how cheap it is.

## Minor

**m1 — `QuestDbDockerFixture.stop()` hands a float to `docker stop -t` (out-of-diff, pre-existing, currently unreachable).** `fixture.py:1072`: `str(wait_timeout_sec)` with a float `120.0` produces `docker stop -t 120.0`, which the docker CLI rejects (`-t` is `ParseInt`). Because `_remove_container` runs with `check=False` then unconditionally `docker rm -f`, the failure is swallowed and the container is hard-killed — silently skipping the graceful WAL-drain the whole bounce design relies on. **This PR does not trigger it** (the fuzz/`BounceThread` path is the only float-passing caller and it skips the docker fixture; all docker-mode `stop()` calls use the int-30 default). It's a genuine latent foot-gun in a file this PR edits, and the trivial fix is `str(int(wait_timeout_sec))`. Optional to fix here since it's out of scope.

**m2 — `_probe_http` `except OSError` misses `http.client.HTTPException` (in-diff, not a regression).** `fixture.py:410`. `BadStatusLine`/`IncompleteRead` are `HTTPException`, not `OSError`, so a malformed/truncated early HTTP response would escape the handler and abort the whole `retry()` loop with a misleading traceback instead of just retrying. Low probability here (the common docker "connect-then-reset" case is `RemoteDisconnected`, which *is* an `OSError` and *is* caught; the managed fixture's port opens with the app). The old code had the same gap, so it's not a regression — but since the helper was just rewritten, adding `http.client.HTTPException` to the except clause is a cheap robustness win.

**m3 — Defensive recovery swallows stop()/start() exceptions with no trace (in-diff).** `qwp_ws_fuzz.py:1536-1545`: both `except Exception: pass` blocks discard the recovery failure. The primary failure is already recorded, but logging the swallowed recovery errors (as CodeRabbit suggested) would help CI diagnosis. Trivial.

**m4 — Managed `start()` doesn't self-clean on failure, unlike docker (in-diff, benign).** `fixture.py:705-708` only prints the log and re-raises, leaving `self._proc` live + `self._log` open; the docker `except` (`:998-1002`) calls `self.stop()`. Verified benign: on the bounce path `BounceThread`'s defensive `stop()` reaps the orphan, and on the initial-start path the caller's `finally: _stop_and_maybe_wipe` (`test.py:4248`) reaps it. Optional symmetry polish — mirror docker with a guarded `self.stop()` — but not a leak.

**m5 — Stuck-bounce-thread wind-down: double-count + teardown race (in-diff, acknowledged, cosmetic).** `test.py:2563-2571`. When the daemon thread is still alive after the join, `record_failure` correctly bumps `failure_counter`, and the assertion at `test.py:2578` short-circuits *before* the verification queries (verified ordering) — so the design goal is met. Residual: the still-alive daemon can race `tearDown`/`stop()` on shared `_proc`/`_log` (e.g. `AttributeError` if one thread nulls `_proc` between another's `if self._proc:` and `.terminate()`), and the failure may be counted twice. Both are cosmetic noise on an already-failed run, and the code comments acknowledge it. Optional: `self.fail(...)` at the detection point to make the abort explicit.

## Nits
- **n1** — `_check_min_http_up` docstring says `HealthCheckProcessor` "answers any path with 200," yet the code probes a specific `/status` and accepts a 2xx *range*. Mildly self-contradictory; align the comment. (Correctness depends on the min server returning 2xx for `/status` on the repo build — the author's smoke test confirms this; if it ever 404'd, every bounce would fail at the 90s cap.)
- **n2** — The budget expression `bounce_stop_timeout_s + BOUNCE_RESTART_TIMEOUT_SEC + 30` is duplicated verbatim in three places (`test.py:2558-2560`, `test.py:2614-2616`, and relied on by `qwp_ws_fuzz.py`). The invariant "client_budget ≥ wind_down ≥ down-window" is held only by copy-paste. Extracting one helper/constant would protect it from silent drift.

## Downgraded (false positives / overstated)
- **"Managed `start()` leaks the JVM on timeout"** (CodeRabbit + agent) — not a leak; the orphan is always reaped by the caller's `finally` (initial start) or `BounceThread`'s defensive `stop()` (bounce). See m4.

## Summary
**Verdict: approve with minor changes.** The fix correctly targets a real, well-diagnosed flake and the timing/attribution redesign is internally consistent and carefully reasoned (the inline comments are excellent). No correctness regressions. The one substantive ask is **M1** (add cheap unit coverage for the probe-selection and status-predicate logic, since a silent predicate flip is the most plausible future regression); everything else is optional polish, with **m1** (docker-stop float) worth a one-character fix while you're in the file even though it's out of scope.

mtopolnik:
## Critical
None.

## Resolution status
_Updated after review fixes landed on `mt_fix-fuzz-timing`. Every change is test infrastructure under `system_test/` / `ci/`; nothing reaches customers._

**✅ Resolved (11):**
- **M1** — `a4753ee4`: Arrow bounce reconnect/close budget is now 270s (120+90+60) and the closing sync uses that finite timeout instead of `0`.
- **m1, m2, n1, n4, n9** — `61c5d996` (Commit A): fixture `stop()`/reap and HTTP-probe hardening.
- **n3, n10** — `be2645d6` (Commit B): column-name out-param retyped to `POINTER(c_void_p)`.
- **m4** — `7302c1f8` (Commit E): `SharedFixtureRecoveryTest` unit-tests both recovery helpers.
- **m6, n7** — `a4753ee4`: `cpp` gate added for the fixture/fuzz suite; docstring corrected.

**⬜ Open / deferred (6):** m3, m5 (Minor); n2, n5, n6, n8 (Nits). Details under each below.

## Moderate

**M1 — The Arrow bounce test lets the server stay down longer than the client is willing to wait**
> ✅ Resolved in `a4753ee4`. The producer's `reconnect_max_duration_millis` / `close_flush_timeout_millis` and the closing sync now derive from `BOUNCE_STOP_TIMEOUT_S + BOUNCE_RESTART_TIMEOUT_S + BOUNCE_LIFECYCLE_HEADROOM_S` (120+90+60 = 270s), and the final drain sync passes that budget instead of `0`.

The new per-bounce limits at `arrow_ingress_fuzz.py:904-905` allow a stop of up to 120s and a restart of up to 90s, so one bounce can leave the server unreachable for up to **210s**. But the producer only waits **120s** to reconnect and to finish its closing flush — `reconnect_max_duration_millis` and `close_flush_timeout_millis`, both `"120000"` at `arrow_ingress_fuzz.py:1407,1409`. If a single stop+restart takes longer than 120s, the producer gives up and the test fails with "producer failed after restart", blaming the client for what was actually a slow server. The QWP fuzz test avoids this by giving its client 270s (120+90+60), which exceeds its own 210s worst case; the Arrow test copied the 120/90 server limits but left the client wait at 120s.
- How likely: **low**. The Arrow bounce test runs one producer sending 5-row batches with no schema changes, so its stops finish in a few seconds and the real down time is almost all restart (~6–40s), well under 120s. It only bites if a stop itself runs for tens of seconds on a slow machine.
- Related: the final drain call `afc.column_sender_sync(conn, 0)` at `arrow_ingress_fuzz.py:1425` passes `0`, which means "wait forever". Against a down server it hangs until the 180s thread-join gives up with a vague "did not finish" message. This PR added a 30s cap (`_SYNC_ON_EXIT_TIMEOUT_MS`) for exactly this reason in `arrow_fuzz_common.py:270`, but not here.
- Fix: compute the Arrow producer's reconnect/close wait from the same `120+90+headroom` the caps use, and give the final sync a real timeout instead of `0`.

## Minor

**m1 — On a failed process kill, `stop()` leaves the log file open and prints no diagnostics**
> ✅ Resolved in `61c5d996` (Commit A). The reap-timeout handler now closes `self._log` and calls `print_log_tail()` before raising `QuestDbReapTimeout`; `_proc` stays set (the tested ownership contract). `test_post_kill_reap_is_bounded_and_preserves_ownership` now asserts the log handle is closed.

When the server can't be killed even after SIGKILL, `stop()` raises `QuestDbReapTimeout` at `fixture.py:927-930` — before the lines that close the log file and clear `_proc` (`fixture.py:931-934`), and before the log-tail dump that `stop()` normally prints on a stuck shutdown (`fixture.py:935` onward). So the worst kind of stop failure prints the least information and leaks the open log handle. Keeping `_proc` set here is correct and is tested; closing the log was just missed. Harmless in the fuzz test (the process exits right after via `os._exit`, which frees the handle), but the Arrow path keeps running.
- Fix: close `self._log` before raising `QuestDbReapTimeout`.

**m2 — `_probe_http` doesn't catch a malformed-response error**
> ✅ Resolved in `61c5d996` (Commit A). Added `import http.client` and widened the handler to `except (OSError, http.client.HTTPException)`. New `test_malformed_response_returns_false` drives a `BadStatusLine` through the probe.

`_probe_http` catches only `OSError` (`fixture.py:432`), but a half-started server can make `urlopen` raise `http.client.BadStatusLine`, which is not an `OSError`. If that happens it escapes `_probe_http`, escapes the `retry()` loop, and makes `start()` fail. Not new (the old code didn't catch it either) and unlikely (a refused, timed-out, or reset connection are all `OSError` and are caught).
- Fix: `except (OSError, http.client.HTTPException)` and add `import http.client`.

**m3 — The isolated-subprocess command only works because of an earlier `sys.argv` rewrite that nothing documents or tests**
> ⬜ Open — deferred (planned as Commit D).

The child command is built from `sys.argv` at `test.py:4403` and injects `--qwp-ws-fuzz-child-dir`, which shares a mutually-exclusive argparse group with `--repo`/`--versions`. This works only because `main()` rewrites `sys.argv` a few lines earlier (`test.py:4641`: `sys.argv[:] = sys.argv[:1] + extra_args`), dropping the already-parsed `run` and `--repo` tokens. Nothing says so near line 4403, `run_with_fixtures` reads the rewritten global instead of the `extra_args` it already holds, and no unit test covers how the child command is assembled or how parent-vs-child suite selection works. A future change that reorders `main()` or calls `run_with_fixtures` directly would break the child command silently, and only the slow integration run would notice.
- To reproduce the dependency: build the child command with and without that rewrite. With the real rewrite the child is `run --qwp-ws-fuzz-child-dir <dir> -v` and parses fine; skip the rewrite and the duplicated `--repo` makes argparse exit 2.
- Fix: build the child arguments from the parsed values (or pass `extra_args` in directly), add a comment, and add a unit test for the assembled command.

**m4 — The Arrow recovery helpers have no unit test**
> ✅ Resolved in `7302c1f8` (Commit E). `SharedFixtureRecoveryTest` in `test_arrow_fuzz_common_unit.py` covers `_restart_if_down` (skip-live / restart-gone / passes `probe_min_http` / swallows a failed start) and `_ensure_fixture_running` (ignore-non-managed / skip-live / reap-then-restart with order asserted / restart-without-reap when `_proc` cleared). Wired into `test.py`'s Arrow import so it runs wherever pyarrow is available.

`arrow_fuzz_common._ensure_fixture_running` and `arrow_ingress_fuzz._restart_if_down` do real work — check whether the server process is still alive, reap a dead one before restarting (`start()` overwrites `_proc`, so skipping the reap would leak the old process), and swallow errors on a best-effort basis. If one regressed, the shared server would quietly stay down and every following SFA test would fail. Integration only exercises the success path.
- Fix: add unit tests for both (the down/reap/restart branches).

**m5 — The QuestDB JVM can be left running with nothing to clean it up, in two cases**
> ⬜ Open — deferred (planned as Commit C; the second sub-point needs a design decision on the `os._exit`-only fallback).

- `run_isolated_suite` sweeps the process tree after a non-zero exit only on Linux/Mac (the `sys.platform != 'win32'` guard). On Windows it relies on `_abort_unrecoverable_lifecycle` killing itself with `taskkill /PID <self> /T /F` (`qwp_ws_fuzz.py:1692`); if that `taskkill` fails, the child still reaches `os._exit(1)` and the supervisor does no Windows sweep, leaving the JVM holding its ports and data dir.
- `_abort_unrecoverable_lifecycle` only kills the process group when the env var `QWP_WS_FUZZ_ISOLATED_CHILD == '1'` (`qwp_ws_fuzz.py:1691`). Run `TestQwpWsFuzz` directly (e.g. `python -m unittest test.TestQwpWsFuzz`) and an unrecoverable failure just calls `os._exit(1)`, orphaning the JVM.

Both are rare or need an unusual way of running the tests.
- Fix: also sweep the tree on Windows for a non-zero exit; and either guard or comment the `os._exit`-only fallback when the env var is unset.

**m6 — The fixture/fuzz unit tests run only on Linux, so their Windows branches never execute**
> ✅ Resolved in `a4753ee4`. `run_python_fixture_tests()` is now gated `if mode in ('all', 'unit', 'cpp')`, so the fixture/fuzz suite also runs on the mac/Windows `cpp` job and its Windows branches execute natively. (The optional `sys.platform`-mocked bounded-wait test on the Linux gate was intentionally not added — mocking the platform on Linux would exercise Python control flow, not the real Windows kill mechanism, which the Windows `integration` run already covers.)

`run_python_fixture_tests()` (which runs `test_qwp_ws_fuzz_unit.py`) is gated to `unit` mode (`ci/run_all_tests.py:156`), and `unit` runs only on the Linux job (`ci/run_tests_pipeline.yaml:51`); the Mac and Windows machines run the split `cargo` + `cpp` jobs and never invoke `unit`. So the `if sys.platform == 'win32'` branches in `IsolatedSuiteSupervisorTest` — the `creationflags=CREATE_NEW_PROCESS_GROUP` check and the `taskkill` bounded-wait check — never run on any machine, reading as coverage that isn't there. The real Windows kill mechanism is still covered by the Windows `integration` run, so this is only about the fast, deterministic unit-level guards. The sibling ctypes suite already got the `'cpp'` gate for this exact reason (`run_all_tests.py:158-161`); the fixture suite didn't.
- Fix: add `'cpp'` to the fixture-tests gate — `if mode in ('all', 'unit', 'cpp'): run_python_fixture_tests()` — so the suite also runs on the mac/Windows `cpp` job and those branches execute natively. The bounded-wait check is also worth running on the fast Linux gate via a `sys.platform` mock (an unbounded `wait()` would hang CI); the `creationflags` assertion just restates the code and isn't worth a mock.

## Nits
- **n1** — ✅ Resolved in `61c5d996` (Commit A). `_check_min_http_up`'s comment says the health server "answers any path with 200," but the code asks for `/status` and accepts any 200–299 (`fixture.py:411-418`). Make the comment match. — _Comment reworded to "probes `/status` and treats any 2xx as healthy."_
- **n2** — ⬜ Open — deferred. The PR description says "both the managed and docker fixtures now discover/publish a 4th port," but there is no docker fixture in `system_test/` — only `QuestDbFixture`, `QuestDbExternalFixture`, and `TlsProxyFixture`. Fix the description.
- **n3** — ✅ Resolved in `be2645d6` (Commit B). The column-name output pointer is still typed `POINTER(_c_char_p)` (`qwp_egress_reader.py:231`), so `name_ptr` is a `c_char_p`, and reading its `.value` would again scan for a NUL past the slice. The fix avoids that by casting; retyping the pointer to `POINTER(_c_void_p)` would remove the trap for good. Not a bug today. — _Retyped to `POINTER(_c_void_p)` and `name_ptr = _c_void_p()`._
- **n4** — ✅ Resolved in `61c5d996` (Commit A). `_probe_http` doesn't close the `urlopen` response (`fixture.py:427`); wrap it in `with urllib.request.urlopen(...) as resp:`. Pre-existing. — _Now wrapped in a `with` block._
- **n5** — ⬜ Open — deferred (fix-or-accept). `run_all_tests.py:29,39` pass exact filenames as the `-p` glob to `unittest discover`. If a file is renamed or misspelled, discover runs 0 tests, which exits 0 (a silent pass) on Python before 3.12.
- **n6** — ⬜ Open — deferred (fix-or-accept). The `qwp_ws_restart` suite still restarts with a plain `QDB_FIXTURE.start()` (main `/ping`, e.g. `test.py` ~2137/2247), keeping the same starvation risk the fuzz suite just moved away from — but it uses a single sender, so the risk is much smaller. Not a regression.
- **n7** — ✅ Resolved in `a4753ee4`. `run_python_fixture_tests`' docstring calls the suite "Pure-Python," but `qwp_ws_fuzz.py:87` imports `questdb_line_sender` at import time, which loads the built library — so the suite actually needs the build. Misleading if a build-free test lane is ever added. — _Docstring now states a prior CMake build is required._
- **n8** — ⬜ Open — deferred (fix-or-accept). `test_qwp_egress_reader_unit.py:214` checks the exact Rust error text; rewording that message on the server side would break the test.
- **n9** — ✅ Resolved in `61c5d996` (Commit A). `QuestDbExternalFixture` never sets `http_min_port` (`fixture.py:534-556`) and the base class doesn't either, while the new `_check_min_http_up` reads it. Can't happen today (external fixtures have no `start()`), but adding `http_min_port = None` on the base class would close the gap. — _Added `http_min_port = None` on `QuestDbFixtureBase`._
- **n10** — ✅ Resolved in `be2645d6` (Commit B). `_collect_column_names` computes `name_addr` for the check but passes `name_ptr` to `string_at` (`qwp_egress_reader.py:920-923`). Both point at the same address, so it's only a cosmetic inconsistency. — _With the void-pointer retype, `name_addr = name_ptr.value` reads the address directly and the cast is gone._

## Summary
**Approve with minor changes.** The main fix is correct and I confirmed it against the QuestDB server source: pointing the readiness probe at the min-HTTP server with its own worker (`http.min.worker.count=1` gives a dedicated pool; `0` would have shared the busy pool and kept the bug). The ctypes fix is correct and complete and has a real regression test. The timing math checks out. Nothing here reaches customers — it's all test code — so no finding is a data-loss risk.

**Status:** the one Moderate (M1) and all Minor/Nit findings except six are now fixed on `mt_fix-fuzz-timing` (commits `a4753ee4`, `61c5d996`, `be2645d6`, `7302c1f8`); all 51 unit tests across the fixture, ctypes-reader, and Arrow suites pass. Still open by choice: m3 and m5 (Minor) and nits n2, n5, n6, n8 — deferred pending a design decision on m5's `os._exit` fallback and fix-or-accept calls on the rest.

mtopolnik:
PR no longer relevant

PR #169, opened 2026-07-17
ci(maven): use the internal Maven cache where reachable, and retry failed downloads

mtopolnik:
## The problem

The hourly fuzz pipeline failed on macOS with:

```
Failed to read artifact descriptor for org.apache.maven:maven-plugin-api:jar:2.0.9
```

Maven was downloading that file from Maven Central over the public internet, and one failed request took the whole run down. Two separate things were wrong.

**Connect-phase download failures were never retried.** Maven does retry a failed download three times out of the box, but both of its HTTP transports pick a retry handler that treats `ConnectException` as non-retryable. A refused or timed-out connection is therefore not retried at all — it fails the build on the first try. Nothing under `ci/` changed that, because no Maven step passed any resolver configuration.

**The jobs on our own machines ignored our Maven cache.** We run an nginx-fronted Reposilite cache at `maven.internal:8081`, reachable only from the Hetzner private network. `TestQwpWsFuzzLinux` and `TestVsQuestDBMaster` run on our `hetzner-incus` machines, which sit on that network — but they were downloading from Maven Central anyway, because nothing had ever told Maven the cache exists.

## What this changes

Adds `ci/templates/maven_settings.yaml`, following the approach in questdb/questdb's `ci/templates/steps.yml`. Before Maven runs, it does two things:

1. **Checks whether `maven.internal:8081` answers.** If it does — meaning we're on a Hetzner machine — Maven is pointed at questdb's `maven-settings-hetzner.xml` and downloads come from the cache. If it doesn't — a Microsoft-provided macOS or Windows machine — Maven gets questdb's plain `maven-settings.xml` and downloads from Maven Central, as it does today.
2. **Sets `MAVEN_RUN_OPTS`,** which makes Maven retry a connect failure instead of giving up on it.

Both settings files live in the questdb repository, so this step has to run after `templates/clone_questdb.yaml`, which is what downloads them.

Seven Maven steps now pass the settings file and the retry options: three jobs in `run_fuzz_pipeline.yaml` and four in `run_tests_pipeline.yaml`.

### Why the retry options are the shape they are

Getting a connect failure retried takes all four of these together, and dropping any one of them makes the other three inert:

- `-Dmaven.resolver.transport=wagon` selects the transport that reads the three `retryHandler` options, and that applies the per-server connect and read timeouts `maven-settings-hetzner.xml` configures. Maven 3.9 otherwise defaults to its `native` transport, which reads neither.
- `-Dmaven.wagon.http.retryHandler.class=default` selects the only handler that honours `nonRetryableClasses`. This is the load-bearing one. Wagon defaults to its `standard` handler, which builds its non-retryable list from a hardcoded set and never reads the option — so without this, the next line has no effect whatsoever.
- `-Dmaven.wagon.http.retryHandler.nonRetryableClasses=java.io.NotSerializableException` replaces the list of exceptions the handler refuses to retry. `ConnectException` is on that list by default; naming an exception that never occurs here leaves the list effectively empty, so a refused or timed-out connection gets retried too.
- `-Dmaven.wagon.http.retryHandler.count=3` is how many attempts a download gets. This restates wagon's default; it is here so the attempt count is visible next to the handler it applies to.

`-Dmaven.wagon.httpconnectionManager.ttlSeconds=30` is separate from retrying: it caps how long a pooled connection may be reused, against a default of 300s. Pooling is on either way; the short cap discards a connection the network may have silently dropped instead of handing it to the next download.

## How it was checked

- **The cache path ran for real on this PR's Hetzner jobs and works.** `Compile QuestDB` on `hetzner-incus` ran with `-s .../maven-settings-hetzner.xml` and pulled 2902 artifacts from `http://maven.internal:8081/releases/`, and zero from Maven Central.
- **The Maven Central path ran on this PR's hosted jobs and works.** macOS and `ubuntu-latest` both logged `Cache at maven.internal is unreachable`, selected the plain settings file, and downloaded `Downloaded from central: https://repo.maven.apache.org/...`.
- **The retry handler was verified against the wagon that ships with Maven, not against the docs.** Invoking wagon's `createRetryHandler()` reflectively and asking the result whether it would retry a `ConnectException`: with `retryHandler.class` left at its default, the handler is `StandardHttpRequestRetryHandler` and the answer is `false`; adding `-Dmaven.wagon.http.retryHandler.class=default` selects the handler that reads `nonRetryableClasses` and the answer becomes `true`. Confirmed on wagon-http 3.5.3 (Maven 3.9.x — the hosted agents) and by bytecode on wagon-http 3.4.3 (the Maven 3.8.x line — Ubuntu's apt Maven on `hetzner-incus`), which have identical branch structure and defaults.
- Looked at what the Hetzner settings file actually does: it redirects every Maven Central request to `maven.internal` and has nothing to fall back on. If a machine that cannot reach the cache were ever given that file, every build on it would fail outright — not occasionally, always. That is why the choice is made by testing reachability on each machine, rather than by a rule about which pool a job runs in.
- Ran the "cannot reach the cache" path: it selects the plain settings file and finishes in roughly 60ms, because the DNS lookup fails long before the 2-second timeout.
- Confirmed that all seven jobs run the steps in the required order — clone the questdb repository, choose the settings file, then run Maven — and that every Maven step passes both the settings file and the retry options.

## What this does not fix

- **On macOS this reduces failures rather than eliminating them.** Retrying helps when a connection fails by chance. If Maven Central genuinely refuses a request, the build still fails. The cache cannot help there: `maven.internal` is reachable only from the Hetzner network, and we have no macOS machines on it.
- **The retry path is configured, not exercised.** No CI run here has actually hit a connect failure and been seen to recover from one. The evidence above is that the handler which retries them is now the one Maven installs.
- **Two settings that questdb uses are deliberately left out.** `MAVEN_KEEPALIVE_OPT` has no effect once `-Dmaven.resolver.transport=wagon` is set, and we always set it. `-Dmaven.repo.local` needs a `MAVEN_CACHE_FOLDER` that these pipelines do not define.
- **This depends on questdb/questdb keeping `ci/maven-settings*.xml` where they are.** If those files are moved or renamed, these pipelines break — Maven exits with `The specified user settings file does not exist` rather than quietly carrying on — with no warning from that side.
- **`ci/run_soak_pipeline.yaml` has the same problem but is not on main yet** — it lives on `jh_conn_pool_refactor`. Its fix travels with that branch, which needs to merge *after* this one. In the other order, main would briefly have a soak pipeline referring to a template that isn't there.

## Note for questdb/questdb

questdb's own `MAVEN_RUN_OPTS` (`ci/test-pipeline.yml`, `ci/test-hosted-pipeline.yml`, `ci/test-fuzz.yml`) omits `retryHandler.class=default`, and the comment at `test-pipeline.yml:21-27` states that `nonRetryableClasses` "re-enables retry on connect failures (the case we actually hit)". It does not — wagon's default `standard` handler never reads that option. The connect-retry mitigation there has been inert since it was added. Filed separately; this PR carries the corrected form.

## Summary by CodeRabbit

* **Build & Reliability**
  * Improved Maven-based build and fuzz-testing pipeline reliability.
  * Builds now automatically select the appropriate Maven configuration based on network availability.
  * Added retry and connection settings to help Maven operations recover from transient repository issues.
  * Applied these improvements across scheduled tests, fuzzing, and live-server validation jobs.


mtopolnik:
## Review: PR #169

**Level 2.** This PR is a CI-configuration change (Azure Pipelines YAML) plus two trivial clippy-idiom fixes in test/dev-tool Rust. There is no production Rust, no FFI, no C/C++ header, and no ILP wire-format surface — so the level-2 Rust/FFI/panic/concurrency agent battery has nothing to bite on. I reviewed the actual change surface directly: the new template's shell/Azure logic, the wiring into all 7 Maven jobs, and the two Rust no-ops.

**Verdict: approve.** Well-constructed, extensively documented, and I could not find a correctness defect. No Critical findings. Everything below is either an acknowledged tradeoff or a nit.

## Moderate

**M1 — New hard dependency on questdb/questdb's `ci/maven-settings*.xml` paths, no fallback (out-of-repo). All 7 Maven jobs now pass `-s $(MAVEN_SETTINGS_FILE)`.** Before this PR, the Maven steps used the agent default (`~/.m2/settings.xml` or none) and depended on nothing in the questdb repo. Now every one of them will exit with `The specified user settings file does not exist` — failing the whole job — if those two files are ever moved or renamed on questdb master. This is a genuine new failure mode spanning both pipelines.

It is, however, **intentional, documented in the PR body, currently working** (the PR verified both files resolve and pull artifacts), mirrors questdb's own approach, and fails *loudly* rather than silently — which is aligned with your fail-fast-over-tolerating-infra preference. Surfacing it only so it's a conscious sign-off, not a blocker.

## Minor

**m1 — The `curl -sf` probe silently demotes a Hetzner machine to Maven Central if the cache answers with any non-2xx.** `-f`/`--fail` makes curl return non-zero on HTTP ≥ 400, so a cache that is *up but degraded* (500s, an auth-gated root, a transient error at `/`) is treated identically to "not on the Hetzner network" → falls back to `maven-settings.xml` → downloads from Central. The build still succeeds (Hetzner reaches Central too), so this is graceful, but a broken cache on a Hetzner machine goes unnoticed except for the one echoed log line — i.e. it *works around* infra misbehavior rather than flagging it, the opposite of the fail-loud stance in M1. This is an unavoidable consequence of using reachability to auto-detect the pool (the probe can't distinguish "not on Hetzner" from "on Hetzner, cache sick"), so it's a reasonable design — just worth being aware the cache can silently stop being used. Secondary nit: the probe hits `/` while artifacts resolve under `/releases/`, so it's testing host-up, not artifact-path-healthy.

**m2 — `-Dmaven.resolver.transport=wagon` is forced on *all* builds, including the fast Hetzner cache path**, opting the whole fleet out of Maven 3.9's default `native` transport. This is deliberate (wagon is the transport that reads the retry handler and the per-server timeouts in `maven-settings-hetzner.xml`) and documented; on the Hetzner apt Maven 3.8.x it's a no-op since wagon is already the default there. No action needed — noting the global scope.

**m3 (nit) — Two unrelated clippy fixes are bundled into a CI PR** (`sender.rs`: `&[b'='][..]` → `&b"="[..]`; `tls_proxy/src/lib.rs`: `{}`,`&dest_addr` → `{dest_addr}`). Both are exact behavioral no-ops (verified below), and both are legitimately needed to keep the `-D warnings` clippy jobs (`run_tests_pipeline.yaml:100,108,121`) green under clippy 1.97 — so the inclusion is defensible. Just flagging the mixed concern.

## Summary
- **Approve.** 1 Moderate (an acknowledged, intentional out-of-repo coupling), 3 Minor (two documented tradeoffs + one nit). No Critical, no correctness bug.

bluestreak01 (approved):

PR #170, opened 2026-07-21
fix(system_test): write-rejection test times out awaiting ack vs QuestDB master

mtopolnik:
## Problem

PR #164's "Vs QuestDB master" CI run failed in
`TestQwpWsProtocol.test_write_rejection_drops_and_sender_continues`:

```
File "system_test/test.py", line 1902, in test_write_rejection_drops_and_sender_continues
    self.assertTrue(sender.await_acked_fsn(final_fsn, 30000))
AssertionError: False is not true
```

The sender never received the final ack within 30s, running against QuestDB
master `8f1e516f9b` (9.4.4-SNAPSHOT, JDK 25). PR #164 only touches
`TestQwpWsFuzz`, so this points at a QuestDB master behavior change (or a new
flake) in how a rejected write interacts with subsequent acks.

## This PR

1. Bumps the `questdb` submodule pointer to current master `f2c567869d`. The
   "Vs QuestDB master" job clones master fresh on every run, so the pointer only
   affects local development — the bump keeps local system-test runs on the same
   code CI exercises and records the commit under investigation.
2. Triggers a fresh CI run to check whether the ack timeout reproduces against
   master tip.
3. Will carry the fix once the failure is understood.


mtopolnik:
Closing without merging. The "Vs QuestDB master" failure this PR reproduced is not a flake or test-infra issue — it is an intentional server-side behavior change. Since questdb/questdb#7359 the server fails fast on deterministic ingest rejections: after rejecting a frame with SCHEMA_MISMATCH it stops processing the connection. Main's client still applies DROP_AND_CONTINUE to that status — it drops the rejected frame and keeps waiting for acks that can no longer arrive, which is exactly the 30s `await_acked_fsn` timeout in `test_write_rejection_drops_and_sender_continues`.

PR #153 already adapts the client to the new contract (NACK policy v2: terminal schema rejections that preserve the store-and-forward data, retriable transient refusals) along with the matching system test, and its "Vs QuestDB master" job is green. A stopgap on main would conflict with #153's rewrite of the same code, so the fix lands there instead.

Until #153 merges, `TestQwpWsProtocol.test_write_rejection_drops_and_sender_continues` will keep failing the "Vs QuestDB master" job on main-based PRs.

PR #173, opened 2026-07-27
ci(maven): rerun the Maven compile task when it fails

mtopolnik:
## What happened

The scheduled fuzz run on 2026-07-27 ([build 255975](https://dev.azure.com/questdb/questdb/_build/results?buildId=255975)) failed on the hosted macOS agent even though the Maven invocation carried the wagon retry options added in #169. The settings-file probe and the retry flags all worked as designed; the agent then lost DNS resolution for `repo.maven.apache.org` for ~30 seconds mid-build. All three wagon retries re-failed instantly out of the JVM's 10-second negative DNS cache without ever touching the network. Back-to-back retries only cover a failure that is over the moment it happens; they cannot outlast an outage.

## The fix

Add `retryCountOnTaskFailure: 2` to every Maven "Compile QuestDB" task across all three pipelines (ten tasks in total: three in `run_fuzz_pipeline.yaml`, six in `run_tests_pipeline.yaml`, one in `run_soak_pipeline.yaml`). A task rerun keeps everything already present in the local Maven repository and spends minutes recompiling the modules built so far before it reaches the download that failed, so it is far more likely to outlast this kind of outage.

This also covers the two integration-test jobs in `run_tests_pipeline.yaml` that never got the wagon retry options, and it applies to the hetzner jobs too, where a rerun rides out a hiccup of the internal cache.

The comment in `templates/maven_settings.yaml` now documents how the two retry layers divide the work.

`retryCountOnTaskFailure` needs agent ≥ 2.194; the hetzner pool runs 4.274.1, so it is honored on every pool.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CI stability by enabling Maven “Compile QuestDB” task retries (up to two) across scheduled fuzz, test, native, and soak pipeline jobs.
  * Fixed native QWP egress reader handling for error messages and column names to safely process non-NUL-terminated strings.
* **Documentation**
  * Clarified Maven settings guidance on when built-in retry handling applies versus when full task retries are needed.

### DIFF
--- a/ci/run_fuzz_pipeline.yaml
+++ b/ci/run_fuzz_pipeline.yaml
@@ -70,6 +70,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -184,6 +185,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -295,6 +297,7 @@ stages:
             displayName: "Install JDK 25"
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               jdkVersionOption: "default"
--- a/ci/run_soak_pipeline.yaml
+++ b/ci/run_soak_pipeline.yaml
@@ -96,6 +96,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
--- a/ci/run_tests_pipeline.yaml
+++ b/ci/run_tests_pipeline.yaml
@@ -58,6 +58,7 @@ stages:
           - template: templates/clone_questdb.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -120,6 +121,7 @@ stages:
           - template: templates/clone_questdb.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -345,6 +347,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -420,6 +423,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -521,6 +525,7 @@ stages:
           - template: templates/maven_settings.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
@@ -581,6 +586,7 @@ stages:
             displayName: "Install JDK 25 (required by QuestDB master)"
           - task: Maven@3
             displayName: "Compile QuestDB"
+            retryCountOnTaskFailure: 2
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               jdkVersionOption: "default"
--- a/ci/templates/maven_settings.yaml
+++ b/ci/templates/maven_settings.yaml
@@ -52,9 +52,12 @@
 # how a scheduled fuzz run on a hosted macOS agent failed despite these
 # options). The Maven compile steps therefore also set
 # retryCountOnTaskFailure: 2, which makes Azure rerun the whole task on
-# failure. A rerun keeps everything already in the local repository and spends
-# minutes recompiling the modules built so far before it reaches the download
-# that failed, so it retries on a timescale that outlasts such an outage.
+# failure. A rerun keeps everything already in the local repository, so a
+# download that failed deep in the build is re-reached only after minutes of
+# recompiling the modules built so far — far more likely to outlast such an
+# outage than a back-to-back retry. A download that fails during the initial
+# dependency resolution is re-reached sooner, so the rerun helps most once the
+# build has made progress.
 #
 # The echo lines run without `set -x`. Azure treats any line containing
 # `##vso[` as a command and takes the text after `]` as the value; tracing the
--- a/ci/templates/maven_settings.yaml
+++ b/ci/templates/maven_settings.yaml
@@ -44,20 +44,12 @@
 #     short cap discards a connection the network may have silently dropped
 #     instead of handing it to the next download.
 #
-# The retryHandler retries fire back-to-back within about a second, so they
-# only cover a failure that is over the moment it happens: one dropped
-# connection, one bad response. A network or DNS outage that lasts longer
-# swallows every attempt (the JVM even caches a failed hostname lookup for 10
-# seconds, so immediate retries re-fail without touching the network — that is
-# how a scheduled fuzz run on a hosted macOS agent failed despite these
-# options). The Maven compile steps therefore also set
-# retryCountOnTaskFailure: 2, which makes Azure rerun the whole task on
-# failure. A rerun keeps everything already in the local repository, so a
-# download that failed deep in the build is re-reached only after minutes of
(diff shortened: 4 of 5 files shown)


mtopolnik:
## Reviewing PR #173 at level 3 — scoped to commit `96082c8` only

Per your instruction, I reviewed **only** `96082c8` ("Rerun the Maven compile task when it fails in CI"); the second commit `e20db9b1` (the cherry-picked `qwp_egress_reader.py` NUL-scan fix) is excluded.

**Scope note on the level-3 machinery:** this commit touches three Azure Pipelines YAML files (21 insertions) — no Rust, C, C++, FFI, wire format, concurrency, allocation, or store-and-forward code. The FFI-safety / panic / Drop / cross-context agent fan-out that level 3 normally drives has no surface to analyze here, so I applied level-3 *thoroughness* to what this change actually is: Azure Pipelines schema validity, byte-exact YAML indentation, repo-wide completeness of coverage, and accuracy of the documentation comment and the factual claims. I verified all of it directly.

**Change surface (semantic delta):** adds `retryCountOnTaskFailure: 2` as a step-level property to nine "Compile QuestDB" `Maven@3` tasks (3 in `run_fuzz_pipeline.yaml`, 6 in `run_tests_pipeline.yaml`), and adds an explanatory comment to the shared `templates/maven_settings.yaml`. No behavioral change to any existing step's inputs.

**What I verified as correct:**
- **Indentation** — `retryCountOnTaskFailure: 2` sits at 12 spaces in every one of the 9 tasks, a proper sibling of `task`/`displayName`/`inputs`, not nested inside `inputs`. Confirmed byte-for-byte from the post-commit tree (`git show 96082c8:…`), not the working branch — note that the current working branch `mt_fix-fuzz-timing` does **not** contain this commit, so the working-tree files are the wrong reference.
- **Schema validity** — `retryCountOnTaskFailure` is a documented step-level `task` property; minimum agent 2.194.0 (matches the PR body); reruns the same task on the same agent so the local `~/.m2` stays warm. Confirmed against Microsoft's official docs.
- **Completeness of the stated scope** — exactly 9 tasks across the two named pipelines; matches "nine tasks in total."
- **Scoping safety** — the retry wraps only the `-DskipTests` compile task; it does **not** wrap any test-execution step, so it cannot mask a flaky test or a QuestDB regression (consistent with the "fail fast on real failures" principle).

### Critical
None.

### Moderate
None.

### Minor

**1. The "outlasts such an outage" rationale is heuristic and depends on where the download fails. — in-diff (comment) / commit message**

The mechanism holds well when the failed download occurs deep in the reactor build: on rerun, the warm `.m2` means Maven spends minutes recompiling earlier modules before it re-reaches the failed download, by which point a ~30s outage is over. But Azure's retry backoff is short — retry #1 after 1s, retry #2 after 4s (per the official docs). If the failure instead happens during up-front plugin/dependency resolution (before meaningful compilation), all three attempts can cluster within a few seconds and may not outlast a 30s outage. The change is still a strict improvement (three independent attempts, each a fresh JVM with a fresh DNS negative-cache, vs. one), so this isn't a defect — but the absolute phrasing "outlasts such an outage" in both the commit message and `maven_settings.yaml:57` is optimistic for the early-failure case. Consider softening to "…so it is far more likely to outlast such an outage."

**2. Retry cost interacts with the tight 30-minute live-server job budget. — out-of-diff**

`ci/run_fuzz_pipeline.yaml` `TestQwpEgressLiveServerFuzz` runs under `timeoutInMinutes: 30` (line 275) and its now-retried compile builds QuestDB master with the web console (`-Pbuild-web-console`), which is not cheap. On a *persistent* compile failure the task now burns up to 3× the compile time before the job fails — which can surface as a job-level timeout rather than a clean compile error, muddying diagnostics. This only bites on genuine persistent failure (where the job fails either way), so it's not blocking, but worth being aware of for that one tight-budget job.

### Downgraded (considered and dismissed)

- **Maven `.lastUpdated` failure-cache defeating the retry** — I checked whether attempt 1 writing a `*.lastUpdated` marker for the failed artifact would make attempt 2 refuse to re-download (Maven's default `daily` `updatePolicy`), silently defeating the whole mechanism. Dismissed: that suppression applies to *not-found* resolutions and snapshot-metadata update intervals, not to transport/DNS `TransferFailedException`s — and the motivating incident was a DNS resolution failure, which a fresh `mvn` invocation re-attempts. (If the team wants belt-and-suspenders, adding `-U` to the compile options would force re-resolution, but it isn't required.)
- **Indentation / schema-placement bug** — none; verified correct in all 9 tasks against the post-commit tree and the official schema.

### Summary

**Verdict: approve.** The nine edits are individually correct — right property, right camelCase spelling, right step-level placement, right indentation, valid on every agent pool in use. The two remaining findings are Minor and non-blocking.

Findings: 2 verified (2 Minor), 1 draft consideration dropped as not-applicable. In-diff vs out-of-diff: Minor #2 is out-of-diff (the tight-budget live-server job), Minor #1 is in-diff (the comment and commit-message wording).

Sources for the schema/agent-version verification:
- [Task types and usage — Azure Pipelines (Microsoft Learn)](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/tasks?view=azure-devops)
- [steps.task schema — retryCountOnTaskFailure](https://github.com/MicrosoftDocs/azure-devops-yaml-schema/blob/main/content/steps-task.md)

mtopolnik:
# PR #173 Review — `ci(maven): rerun the Maven compile task when it fails`

**Level 3.** This PR is two things: CI pipeline config (YAML), and a ~20-line fix to the Python test harness that reads data back from the Rust library (`system_test/qwp_egress_reader.py`). None of the actual shipping library changes here — no Rust, C, C++, wire format, threading, or memory-management code — so there's little for the usual deep level-3 checks to examine. I still reviewed carefully: I checked the Python change against what the Rust functions actually promise on both sides of the boundary, had a second reviewer run ctypes directly to confirm how it behaves, and checked the CI YAML for valid syntax, correct indentation, complete coverage across the repo, and whether the PR description matches what changed.

### What changed
- **CI YAML:** adds `retryCountOnTaskFailure: 2` to **all 10** "Compile QuestDB" Maven tasks across three pipeline files (fuzz ×3, soak ×1, tests ×6). It's a property on the task itself, so Azure reruns the whole task if it fails. Nothing about what each task does changes. Plus an explanatory comment in `templates/maven_settings.yaml`.
- **`qwp_egress_reader.py`:** The Rust FFI hands back error messages and column names as a pointer plus a separate length — the bytes are *not* terminated by a NUL, and the string is only borrowed (valid until the next batch). The old code told ctypes the return was a C string (`c_char_p`), so ctypes scanned forward looking for a NUL to mark the end. Since there is no NUL, that scan runs past the end of the string into whatever memory follows. The fix reads the pointer as a plain address and copies exactly the number of bytes the length says. Same decoded result on the normal path; the difference is it no longer reads past the end.

## Critical
None.

## Moderate
None.

## Minor

**1. The PR description is out of date — it says "both pipelines (nine tasks)", but it's now three pipelines and ten tasks. — description**
The description was written for the first commit (fuzz 3 + tests 6 = 9 tasks, two pipelines). A later commit (`b602f2c5`) added a 10th task in a third pipeline, `run_soak_pipeline.yaml`, and the description wasn't updated. A repo-wide search confirms 10 occurrences: fuzz (3), soak (1), tests (6). Please update the description to "three pipelines / ten tasks."

> ✅ **Done.** The description now reads "all three pipelines (ten tasks in total: three in `run_fuzz_pipeline.yaml`, six in `run_tests_pipeline.yaml`, one in `run_soak_pipeline.yaml`)".

**2. "outlasts such an outage" is more confident than the mechanism guarantees. — `maven_settings.yaml:47-57` and the commit message**
The comment explains that a rerun *"spends minutes recompiling the modules built so far before it reaches the download that failed."* That's true only when the download fails deep into the build. If a download fails early — during the initial dependency and plugin resolution, before much has compiled — the rerun gets back to that same download within seconds, and Azure waits only a short time between attempts. In that case all three attempts can happen inside a ~30-second outage and all fail. The change is still a clear improvement (three separate attempts, each a fresh JVM that re-resolves DNS), so this isn't a bug — but consider changing "so it retries on a timescale that **outlasts** such an outage" to "…so it is **far more likely to outlast**…".

> ✅ **Done** (commit `0ba60cb0`). The comment now reads "far more likely to outlast such an outage than a back-to-back retry" and adds the early-failure caveat: "A download that fails during the initial dependency resolution is re-reached sooner, so the rerun helps most once the build has made progress." The PR description carries the matching wording.

**3. On the 30-minute live-server fuzz job, a real compile failure can now show up as a timeout instead of a clean error. — `run_fuzz_pipeline.yaml:275-300`**
`TestQwpEgressLiveServerFuzz` has a 30-minute limit (line 279), and its compile task (line 300) builds QuestDB master. Azure retries on *any* task failure, so if the compile fails for a real, repeatable reason, it now runs up to three times before the job gives up — which can push the job past 30 minutes and report a timeout rather than "compile failed", making it harder to see what actually went wrong. The other three fuzz jobs have 90-minute limits and aren't affected. This only happens when the build genuinely can't compile (the job fails either way), so it doesn't block merge.

> ⏭️ **Won't fix (by decision).** This only manifests when the build genuinely can't compile — the job fails either way — and it's non-blocking as noted. Not worth the added complexity of exempting this one job from the retry or bumping its limit.

## Summary
**Verdict: approve.** The Python change is a real fix: it stops the harness from reading past the end of a borrowed, non-terminated string, which could crash if that string sits at the edge of a memory page. It also makes the reader handle error strings the same way the sender already does. The 10 CI edits are each correct — right property, right spelling, right placement and indentation, complete coverage, valid on every agent pool — and they only wrap the compile step, so they can't mask a failing test or a product regression.

Two small things worth doing before merge, neither required: update the description's "nine tasks / both pipelines" to "ten tasks / three pipelines", and optionally soften the "outlasts such an outage" wording in `maven_settings.yaml` and the commit message.

---

**Resolution:** m1 ✅ done · m2 ✅ done (`0ba60cb0`) · m3 ⏭️ won't fix (non-blocking, by decision).

jerrinot (approved):

PR #174, opened 2026-07-27
docs(qwp): define standalone direct sender drop contract

mtopolnik:
## Summary

- clarify that `qwp_direct_sender` handles may be pool-borrowed or standalone
- define the normal-release and force-drop contract for both ownership modes
- explicitly support `questdb_db_drop_direct_sender(NULL, sender)` for standalone failure cleanup
- mirror the public contract in the Rust FFI docs and pin it with documentation and ABI regression checks

## Context

A client review flagged the NULL `db` call as a possible crash or leak. The Rust implementation already ignores that argument and retains the pool-versus-standalone backing in the sender itself, so the runtime behavior is safe. The public C documentation, however, only described borrowed handles and did not exempt `db` from the general non-NULL opaque-handle convention.

This PR aligns the existing API contract with the implemented ownership semantics. It does not change ABI signatures or add a new symbol.

## Testing

- `python3 ci/check_docs.py`
- `cargo fmt --manifest-path questdb-rs-ffi/Cargo.toml -- --check`
- `cargo test --manifest-path questdb-rs-ffi/Cargo.toml` — 87 passed
- `cargo clippy --manifest-path questdb-rs-ffi/Cargo.toml --tests`

## Summary by CodeRabbit

- **Documentation**
  - Expanded direct sender lifecycle and failure-handling guidance, including when to return, free, or force-drop handles.
  - Clarified standalone vs pooled handle expectations and the correct cleanup after failures, including `db` being allowed to be `NULL`.
  - Documented connection-close behavior and the mutually exclusive release/discard methods.

- **Tests**
  - Added integration coverage verifying standalone direct-sender drop behavior for deferred frames: best-effort commit on the “free” path vs discarding uncommitted deferred frames on the “force drop” path.
  - Added checks that discarding accepts `NULL` for both `db` and the sender handle.

### DIFF
--- a/ci/check_docs.py
+++ b/ci/check_docs.py
@@ -380,6 +380,10 @@ def check_compatibility(errors):
                 "Enterprise-only `qwpws_ack_level_durable`",
                 "durable ACK edition qualifier",
             ),
+            (
+                "NULL for a standalone handle",
+                "standalone direct-sender force-drop contract",
+            ),
         ),
         ROOT / "include" / "questdb" / "ingress" / "line_sender.h": (
             (
--- a/ci/check_docs.py
+++ b/ci/check_docs.py
@@ -399,6 +399,10 @@ def check_compatibility(errors):
                 "Enterprise-only\n * `qwpws_ack_level_durable`",
                 "durable ACK edition qualifier",
             ),
+            (
+                "For a standalone handle, pass NULL for `db`.",
+                "standalone direct-sender force-drop contract",
+            ),
         ),
     }
     for path, required_text in public_contract_requirements.items():
--- a/include/questdb/ingress/qwp_sender.h
+++ b/include/questdb/ingress/qwp_sender.h
@@ -89,16 +89,17 @@ extern "C" {
  *  store-and-forward queue owns delivery. */
 typedef struct qwp_sender qwp_sender;
 
-/** Borrowed direct (pipelined, non-store-and-forward) column-major QWP/WS
- *  sender from the always-direct pool, independent of `sf_dir`. Not
- *  thread-safe; belongs to the borrowing thread until returned via
- *  `questdb_db_return_direct_sender`.
+/** Direct (pipelined, non-store-and-forward) column-major QWP/WS sender.
+ *  Either borrowed from the always-direct pool, independent of `sf_dir`, or
+ *  opened standalone by `qwp_direct_sender_from_conf` /
+ *  `qwp_direct_sender_from_opts`. Not thread-safe; belongs to the calling
+ *  thread until released through the matching return, drop, or free function.
  *
  *  Exposes publish-only `qwp_direct_sender_flush`, the
  *  `qwp_direct_sender_commit` boundary, and the Arrow batch variants.
  *  There is no internal failover: on a transient
- *  `line_sender_error_failover_retry` the caller drops the sender, re-borrows
- *  a live one, and re-drives the uncommitted tail from its own source. */
+ *  `line_sender_error_failover_retry` the caller drops the sender, obtains a
+ *  new one, and re-drives the uncommitted tail from its own source. */
 typedef struct qwp_direct_sender qwp_direct_sender;
 
 /** One DataFrame's worth of column buffers destined for one QuestDB table.
@@ -251,8 +252,10 @@ qwp_direct_sender* questdb_db_borrow_direct_sender_with_retry(
  * QWP/WebSocket config string, owning its own connection with no pool — for
  * one-off DataFrame bulk loads without a `questdb_db`. `conf` is a UTF-8
  * string of `conf_len` bytes. Returns NULL on failure and sets `*err_out` if
- * provided. Free the returned handle with `qwp_direct_sender_free` (there
- * is no pool to return it to).
+ * provided. On normal completion free the returned handle with
+ * `qwp_direct_sender_free` (there is no pool to return it to). To discard
+ * uncommitted frames after a failure, call
+ * `questdb_db_drop_direct_sender(NULL, sender)` instead.
  */
 QUESTDB_CLIENT_API
 qwp_direct_sender* qwp_direct_sender_from_conf(
@@ -267,8 +270,10 @@ qwp_direct_sender* qwp_direct_sender_from_conf(
  * the `line_sender_opts_*` builder functions rather than a config string — so
  * this works for a sender configured either way. `opts` is borrowed, not
  * consumed: the caller retains ownership and must still free it. Returns NULL
- * on failure and sets `*err_out` if provided. Free the returned handle with
- * `qwp_direct_sender_free` (there is no pool to return it to).
+ * on failure and sets `*err_out` if provided. On normal completion free the
+ * returned handle with `qwp_direct_sender_free` (there is no pool to return it
+ * to). To discard uncommitted frames after a failure, call
+ * `questdb_db_drop_direct_sender(NULL, sender)` instead.
  */
 QUESTDB_CLIENT_API
 qwp_direct_sender* qwp_direct_sender_from_opts(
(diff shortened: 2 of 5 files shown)


mtopolnik:
# Review: PR #174

## ✅ Update — all findings addressed

Follow-up commits on `mt_fix-ffi-contract` resolve every finding below (no Critical were found). Nothing outstanding.

| Finding | Status | Commit |
|---|---|---|
| **M1** — no positive-path test for the standalone contract | ✅ Addressed | 681770cd |
| **m1** — "accepts only" over-states the impl | ✅ Addressed | 29e8c04b |
| **m2** — "later borrower" rationale reads as universal | ✅ Addressed | 29e8c04b |
| **m3** — stale Python `qwp_direct_sender*` docstring | ✅ Addressed | 29e8c04b |
| **m4** — CI pin guarded only the C header | ✅ Addressed | 29e8c04b |

**681770cd — Add standalone direct-sender drop/free tests.** Two `ffi-support` tests in `questdb-rs/src/tests/qwp_sender_pool.rs` build a real standalone `OwnedDirectColumnSender` against the QWP/WS mock and assert the commit-vs-discard split on the wire: a plain drop (the `qwp_direct_sender_free` path) closes the deferred tail with a commit boundary; a `mark_must_close` drop (the `questdb_db_drop_direct_sender(NULL, …)` path) leaves it uncommitted/discarded. The two share identical setup and differ only by the `must_close` mark, asserting opposite commit-flag outcomes, so neither passes vacuously. Covers `from_conf`/`from_builder`, `commit_in_flight_on_drop`, and the standalone `Drop` arm.

**29e8c04b — Tighten standalone direct-sender docs and CI pin.** `questdb_db_return_direct_sender` "accepts only" → "is intended only for"; the drop "later borrower" rationale scoped to pooled senders; the Python docstring → "pool-borrowed or standalone"; and a parallel Rust-surface pin added to `ci/check_docs.py`.

---

*Original level-3 review below, kept for the record.*

**Reviewed at level 3.** This diff contains **zero executable logic** — documentation across `qwp_sender.h` and `column_sender.rs`, one new CI required-string in `check_docs.py`, and one trivial test. The entire review therefore reduces to a single question: *does the pre-existing implementation actually honor every sentence of the newly-written contract?* I verified that end-to-end by reading the implementation, then confirmed it with two independent agents (including the mandatory fresh-context adversarial pass). The structured correctness/safety/concurrency fan-out was collapsed into direct source tracing because there is no new code for those lenses to analyze.

## Critical
None. The documented contract is accurate and the release machinery is memory-safe on all documented single-release paths.

## Moderate

**M1 — The newly-formalized standalone contract has no positive-path test; the one added test only hits the NULL guard.** *(in-diff + out-of-diff)*

> ✅ **Addressed in 681770cd** — two standalone tests now exercise the positive paths (free-commits and force-drop-discards) against the QWP/WS mock, covering `from_conf`/`from_builder`, `commit_in_flight_on_drop`, and the standalone `Drop` arm. See the summary at the top.

`questdb_db_drop_direct_sender_accepts_null_db_and_sender` (`questdb-rs-ffi/src/column_sender.rs:6050-6055`) passes a NULL `sender`, so `return_or_drop_cs` returns at its `if sender.is_null() { return; }` guard (`column_sender.rs:1155-1157`) **before** any backing-release logic runs. It is behaviorally a duplicate of the pre-existing `qwp_direct_sender_free_accepts_null` (`:6045`), and because `_db` is `_`-prefixed/unused, "accepts NULL db" is a compile-time triviality. None of the contract's substantive promises are exercised anywhere:
- `questdb_db_drop_direct_sender(NULL, real_standalone)` discarding uncommitted frames / "closed outright" — untested.
- `qwp_direct_sender_free(standalone)` committing best-effort — untested.
- The core `OwnedDirectColumnSender` standalone Drop commit-vs-discard branch (`db.rs:2345-2349`) and `commit_in_flight_on_drop` (`db.rs:2884`) — never exercised; `from_conf`/`from_builder` (`db.rs:2279`,`2287`) have no test caller (only the compile-time `assert_send` at `db.rs:3251`).

This is a coverage gap, not a code bug — the behavior is verified-correct by reading. But the PR's entire purpose is to *promise this contract to external C callers*, and there is no regression guard for the discard-vs-commit distinction. The gap is closeable: the mock-server infrastructure already exists (`cpp_test/qwp_mock_server.cpp`, `system_test/fixture.py`) but isn't wired to the direct sender. **Recommendation:** add one end-to-end test against the live/mock fixture — build a standalone sender via `from_conf`, stage frames, then assert rows are absent after `questdb_db_drop_direct_sender(NULL, s)` and present after a committed `qwp_direct_sender_free(s)`. (CodeRabbit's inline comment flags the same.)

## Minor

**m1 — "accepts only handles borrowed through…" over-states the implementation.** *(in-diff)*

> ✅ **Addressed in 29e8c04b** — reworded to "is intended only for handles borrowed through…", so it reads as guidance rather than a false rejection claim.

`qwp_sender.h:288-290` and `column_sender.rs:1167-1169` say `questdb_db_return_direct_sender` "accepts only" pool-borrowed handles and standalone handles "must" use `qwp_direct_sender_free`. Both functions are the *identical* call `return_or_drop_cs(sender, 0)` (`:1171-1176` vs `:1218-1220`), so a standalone handle passed to `return_direct` is silently accepted and released correctly. Defensible as a forward-looking prescriptive precondition (reserving future divergence), but as worded it asserts an acceptance restriction the code does not implement. Consider "intended only for…" rather than "accepts only…".

**m2 — The force-drop rationale is pool-specific but reads as universal.** *(in-diff)*

> ✅ **Addressed in 29e8c04b** — the "later borrower" clause now reads "for a pooled sender this stops a later borrower from…", scoping the hazard to the pooled case.

In the `questdb_db_drop_direct_sender` block, *"otherwise a later borrower could commit those frames together with its own batch"* only applies to pooled handles — a standalone connection has no later borrower. The following sentence ("a standalone sender is closed outright") clarifies, so no careful reader is misled, but the "later borrower" motivation could note it is the pooled hazard specifically.

**m3 — Python binding docstring is now incomplete.** *(out-of-diff)*

> ✅ **Addressed in 29e8c04b** — docstring now reads `"(pipelined connection: pool-borrowed or standalone)"`.

`system_test/arrow_ffi.py:39` documents `qwp_direct_sender*` as only *"(borrowed pipelined connection)"*. This PR broadens the type to also cover standalone handles, rendering the docstring incomplete (though accurate for Python's borrow-only usage, hence Minor). Not touched by the diff, but exposed by the contract change.

**m4 — CI pin guards only the C header, not the Rust docs.** *(in-diff)*

> ✅ **Addressed in 29e8c04b** — added a parallel pin on the Rust surface (anchor `"NULL for a standalone handle"` in `questdb-rs-ffi/src/column_sender.rs`).

The new `check_docs.py:402-405` required string (`"For a standalone handle, pass NULL for \`db\`."`) is bound only to `qwp_sender.h`. The equivalent Rust doc comments (`column_sender.rs:1194-1200`) carry the same contract but nothing guards against their silent removal/drift. Optional: add a parallel pin for the Rust surface.

## Summary

**Verdict: Approve.** The documentation is accurate — every claim (db-ignored/NULL-safe, drop-discards, free/return-commits-best-effort, pooled-drop-closes-not-recycles, standalone-closed-outright, double-release-is-UB) matches the implementation, the CI docs gate passes (`python3 ci/check_docs.py` → exit 0), the new exemption is properly reconciled with the header's global "handles must be non-NULL unless documented otherwise" convention (`qwp_sender.h:39-40`), and the Rust and C surfaces are mutually consistent. The only substantive gap is **M1** (no positive-path test for the contract being formalized) — worth addressing before merge but not a correctness blocker for a docs change.

**Update:** all findings (1 Moderate + 4 Minor; no Critical) are now addressed in 681770cd and 29e8c04b — see the status table at the top.

jerrinot (approved):

PR #189, opened 2026-08-20
fix(ingress): publish the dispatcher close under the inbox mutex

mtopolnik:
## The bug

`EventDispatcher::drop` published the close outside the inbox mutex:

```rust
self.inner.closed.store(true, Ordering::Release);
self.inner.available.notify_one();
... handle.join();
```

`dispatch_loop` reads `closed` and parks on that same condvar *while holding the inbox mutex*. Nothing serialises the two, so a drop that lands between the loop's read of `closed` and its park signals zero waiters. `offer` discards events once `closed` is set, so no later signal can arrive — the thread parks for good and the `join` in `drop` blocks forever.

The window is widest right after construction, when the freshly spawned thread makes its first trip through the lock. A dispatcher created and dropped back to back is therefore the most exposed shape.

## Why it matters

Every `questdb_db_connect_with_handlers` installs a rejection handler, so **every pool spawns this thread and joins it in `questdb_db_close`**. Any short-lived pool can stall on close.

It surfaced as a py-questdb-client CI timeout ([Azure build 262974](https://dev.azure.com/questdb/6c9c1a0a-74cf-4f7b-bf65-24ae4f3cd61d/_build/results?buildId=262974), `cibuildwheel macos_cp310_cp311`): a test whose client raises during argument validation and closes without ever connecting hit the 15-minute watchdog. Sampled stacks show both ends exactly:

```
main:                QuestDB.close -> questdb_db_close -> QuestDb::drop
                     -> RejectionEventSource::close -> EventDispatcher::shutdown
                     -> pthread_join                              [blocked]
questdb-rejections:  dispatch_loop -> __psynch_cvwait             [blocked]
```

Looping that client shape locally stalled after 76 / 109 / 150 iterations across three trials.

## The fix

Publish the close under the inbox mutex — the same thing `QuestDb::drop` already does deliberately for the pool condvar (`db.rs`, "Notifying under the mutex avoids the lost-wakeup race"). The dispatch loop is then left with only two possibilities: observe the store before parking, or be registered as a waiter when the notify lands.

No self-deadlock risk: `dispatch_loop` releases the inbox guard before invoking the listener, so a listener that drops the dispatcher is not already holding it.

## Test

`drop_joins_a_dispatcher_thread_that_never_saw_an_event` sweeps the gap between construction and drop across the startup window. It runs the cycle on a worker thread and waits on a channel, so a stall is reported as a receive timeout instead of hanging the suite.

- Without the fix: **fails** (`a dispatcher drop stalled joining its thread`).
- With the fix: **passes** in ~0.6 s, stable over 5 consecutive runs.

## Verification

- `cargo test` in `questdb-rs` — 1690 passed, 0 failed, 23 ignored (plus all doc/integration targets green)
- `cargo test` in `questdb-rs-ffi` — 88 passed, 0 failed
- `cargo fmt` applied; `cargo clippy --tests` clean

## Note

I audited the other condvar sites in `db.rs` and `qwp_ws.rs` for the same pattern; this was the only instance.

- `BackpressureNotifier::notify_all` bumps `generation` **under** its own mutex, and `wait_for_change` reads it under that same mutex — correctly guarded. Every `backpressure.notify_all()` call in `qwp_ws.rs` goes through that method.
- The pool sites in `db.rs` (`InUseSlot::drop`, `SenderSlotRelease::drop`, `finish_*_sender`, reader release, SF recovery) mutate the mutex-protected `state` before notifying, so the mutex serialises them against a waiter that has not parked yet.
- `QuestDb::drop` sets `shutdown` outside the mutex but issues the notify **under** it, which closes the same window from the other side.
- Apart from `BackpressureNotifier::wait_for_change`, every remaining condvar wait in both files is a `wait_timeout`, so a missed wakeup there would cost latency rather than stall.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of event-dispatcher shutdown and repeated close operations.
  - Prevented shutdown workflows from hanging while dispatchers start or stop.
  - Added stronger safeguards around dispatcher startup failures and worker termination.

- **Tests**
  - Expanded concurrency and shutdown coverage across thousands of repeated cycles.
  - Added clearer detection of startup timeouts and worker failures.
  - Preserved existing hashing, encoding, decoding, and data-transfer behavior.

### DIFF
--- a/questdb-rs-ffi/src/lib.rs
+++ b/questdb-rs-ffi/src/lib.rs
@@ -5368,9 +5368,9 @@ mod tests {
         padded.extend_from_slice(&bit_len.to_be_bytes());
 
         let mut words = [0u32; 80];
-        for chunk in padded.chunks_exact(64) {
-            for (idx, word) in chunk.chunks_exact(4).enumerate() {
-                words[idx] = u32::from_be_bytes([word[0], word[1], word[2], word[3]]);
+        for chunk in padded.as_chunks::<64>().0 {
+            for (idx, word) in chunk.as_chunks::<4>().0.iter().enumerate() {
+                words[idx] = u32::from_be_bytes(*word);
             }
             for idx in 16..80 {
                 words[idx] = (words[idx - 3] ^ words[idx - 8] ^ words[idx - 14] ^ words[idx - 16])
--- a/questdb-rs/src/egress/decoder.rs
+++ b/questdb-rs/src/egress/decoder.rs
@@ -1138,7 +1138,7 @@ fn reverse_uuid_rows(buf: ColumnBuffer) -> ColumnBuffer {
         Ok(values) => values,
         Err(shared) => BytesMut::from(&shared[..]),
     };
-    for row in values.chunks_exact_mut(16) {
+    for row in values.as_chunks_mut::<16>().0 {
         row.reverse();
     }
     ColumnBuffer {
@@ -1743,13 +1743,13 @@ fn count_nulls(bitmap: &[u8], row_count: usize) -> usize {
     // bit-scan to find the *position* of a null), switch to
     // `from_le_bytes` — positions are endian-sensitive.
     let body = &bitmap[..full_bytes];
-    let mut chunks = body.chunks_exact(8);
+    let (words, tail) = body.as_chunks::<8>();
     let mut nulls: usize = 0;
-    for c in chunks.by_ref() {
-        let w = u64::from_ne_bytes(c.try_into().unwrap());
+    for c in words {
+        let w = u64::from_ne_bytes(*c);
         nulls += w.count_ones() as usize;
     }
-    for b in chunks.remainder() {
+    for b in tail {
         nulls += b.count_ones() as usize;
     }
     if tail_bits != 0 {
@@ -2309,7 +2309,7 @@ mod tests {
         });
 
         let mut expected = input;
-        for row in expected.chunks_exact_mut(16) {
+        for row in expected.as_chunks_mut::<16>().0 {
             row.reverse();
         }
         assert_eq!(reversed.values, expected);
@@ -2332,7 +2332,7 @@ mod tests {
         });
 
         let mut expected = input.clone();
-        for row in expected.chunks_exact_mut(16) {
+        for row in expected.as_chunks_mut::<16>().0 {
             row.reverse();
         }
         assert_eq!(reversed.values, expected);
--- a/questdb-rs/src/ingress/column_sender/arrow_batch.rs
+++ b/questdb-rs/src/ingress/column_sender/arrow_batch.rs
@@ -925,9 +925,14 @@ fn write_qwp_bitmap_from_arrow(out: &mut Vec<u8>, nulls: &NullBuffer) -> Result<
         let word_bytes = (full_bytes / 8) * 8;
         let (src_words, src_rem) = src_slice.split_at(word_bytes);
         let (dst_words, dst_rem) = dst_slice.split_at_mut(word_bytes);
-        for (dchunk, schunk) in dst_words.chunks_exact_mut(8).zip(src_words.chunks_exact(8)) {
-            let w = u64::from_ne_bytes(schunk.try_into().unwrap());
-            dchunk.copy_from_slice(&(!w).to_ne_bytes());
+        for (dchunk, schunk) in dst_words
+            .as_chunks_mut::<8>()
+            .0
+            .iter_mut()
+            .zip(src_words.as_chunks::<8>().0)
+        {
+            let w = u64::from_ne_bytes(*schunk);
+            *dchunk = (!w).to_ne_bytes();
         }
         for (d, &s) in dst_rem.iter_mut().zip(src_rem) {
             *d = !s;
@@ -1193,7 +1198,7 @@ fn write_uuid_be_payload(out: &mut Vec<u8>, arr: &FixedSizeBinaryArray) -> Resul
             // Bulk path: slice `value_data()` once, starting at the array's
             // own offset, instead of a bounds-checked `value(row)` per row.
             let start = arr.offset() * 16;
-            for row in arr.value_data()[start..start + bytes].chunks_exact(16) {
+            for row in arr.value_data()[start..start + bytes].as_chunks::<16>().0 {
                 out.extend_from_slice(&reverse_uuid_bytes(row));
             }
         }
--- a/questdb-rs/src/ingress/column_sender/conn.rs
+++ b/questdb-rs/src/ingress/column_sender/conn.rs
@@ -1161,24 +1161,18 @@ fn write_ws_header(out: &mut [u8], payload_len: usize, mask_key: [u8; 4]) {
     const BINARY_OPCODE: u8 = 0x2;
     const MASK_BIT: u8 = 0x80;
     out[0] = FIN_BIT | BINARY_OPCODE;
-    let len_bytes;
-    let mask_offset;
-    if payload_len <= 125 {
+    let mask_offset = if payload_len <= 125 {
         out[1] = MASK_BIT | (payload_len as u8);
-        mask_offset = 2;
-        len_bytes = 0;
+        2
     } else if payload_len <= 0xFFFF {
         out[1] = MASK_BIT | 126;
         out[2..4].copy_from_slice(&(payload_len as u16).to_be_bytes());
-        mask_offset = 4;
-        len_bytes = 2;
+        4
     } else {
         out[1] = MASK_BIT | 127;
         out[2..10].copy_from_slice(&(payload_len as u64).to_be_bytes());
-        mask_offset = 10;
-        len_bytes = 8;
-    }
-    let _ = len_bytes;
+        10
+    };
     out[mask_offset..mask_offset + 4].copy_from_slice(&mask_key);
 }
 
--- a/questdb-rs/src/ingress/column_sender/numpy_wire.rs
+++ b/questdb-rs/src/ingress/column_sender/numpy_wire.rs
@@ -2166,7 +2166,7 @@ mod tests {
         // column body subsequence appears exactly once.
         let mut body: Vec<u8> = Vec::new();
         body.push(0u8); // null_flag = 0 (no validity)
-        for row_chunk in rows.chunks_exact(3) {
+        for row_chunk in rows.as_chunks::<3>().0 {
             body.push(1u8); // ndim
             body.extend_from_slice(&3u32.to_le_bytes()); // dim
             for &v in row_chunk {
--- a/questdb-rs/src/ingress/conn_events.rs
+++ b/questdb-rs/src/ingress/conn_events.rs
@@ -131,6 +131,10 @@ struct DispatcherInner<T> {
     closed: AtomicBool,
     dropped: AtomicU64,
     delivered: AtomicU64,
+    /// Set by the worker as it enters `dispatch_loop`, so a test can tell
+    /// a running worker from one that never spawned.
+    #[cfg(test)]
+    worker_started: (Mutex<bool>, Condvar),
 }
 
 impl<T> DispatcherInner<T> {
@@ -139,6 +143,38 @@ impl<T> DispatcherInner<T> {
             Ok(inbox) => inbox,
             Err(poisoned) => poisoned.into_inner(),
         }
+    }
+
+    #[cfg(test)]
+    fn signal_worker_start(&self) {
+        let (started, signal) = &self.worker_started;
+        let mut started = match started.lock() {
+            Ok(started) => started,
+            Err(poisoned) => poisoned.into_inner(),
+        };
+        *started = true;
+        signal.notify_all();
+    }
+
+    #[cfg(test)]
+    fn wait_for_worker_start(&self, timeout: std::time::Duration) -> bool {
+        let (started, signal) = &self.worker_started;
+        let deadline = std::time::Instant::now() + timeout;
+        let mut started = match started.lock() {
+            Ok(started) => started,
+            Err(poisoned) => poisoned.into_inner(),
+        };
+        while !*started {
+            let now = std::time::Instant::now();
+            if now >= deadline {
+                return false;
+            }
+            started = match signal.wait_timeout(started, deadline - now) {
+                Ok((started, _)) => started,
+                Err(poisoned) => poisoned.into_inner().0,
+            };
+        }
+        true
     }
 }
 
@@ -185,6 +221,8 @@ impl<T: std::fmt::Debug + Send + 'static> EventDispatcher<T> {
             closed: AtomicBool::new(false),
             dropped: AtomicU64::new(0),
             delivered: AtomicU64::new(0),
+            #[cfg(test)]
+            worker_started: (Mutex::new(false), Condvar::new()),
         });
         let thread_inner = Arc::clone(&inner);
         let thread = match std::thread::Builder::new()
@@ -231,6 +269,15 @@ impl<T: std::fmt::Debug + Send + 'static> EventDispatcher<T> {
         self.inner.delivered.load(Ordering::Relaxed)
     }
 
+    /// Test-only: block until the worker thread has entered its dispatch
+    /// loop, returning `false` on timeout or when the spawn failed. Lets a
+    /// test that exercises the close/join path prove there was a worker to
+    /// join, rather than passing because no thread was ever created.
+    #[cfg(test)]
+    pub(crate) fn wait_for_worker_start(&self, timeout: std::time::Duration) -> bool {
+        self.thread.is_some() && self.inner.wait_for_worker_start(timeout)
+    }
+
     /// Drop the dispatcher (joining its thread, so any in-flight listener
     /// invocation completes) and return the final `(delivered, dropped)`.
     pub(crate) fn shutdown(self) -> (u64, u64) {
@@ -275,6 +322,8 @@ impl<T> Drop for EventDispatcher<T> {
 }
 
 fn dispatch_loop<T: std::fmt::Debug>(inner: Arc<DispatcherInner<T>>) {
+    #[cfg(test)]
+    inner.signal_worker_start();
     loop {
         let event = {
             let mut inbox = inner.lock_inbox();
--- a/questdb-rs/src/ingress/conn_events.rs
+++ b/questdb-rs/src/ingress/conn_events.rs
@@ -245,8 +245,18 @@ impl<T: std::fmt::Debug + Send + 'static> EventDispatcher<T> {
 
 impl<T> Drop for EventDispatcher<T> {
     fn drop(&mut self) {
-        self.inner.closed.store(true, Ordering::Release);
-        self.inner.available.notify_one();
+        // The inbox mutex is required for the wakeup to reach the
+        // dispatcher: `dispatch_loop` reads `closed` and parks while
+        // holding that same mutex, so publishing the close under it means
+        // the loop either sees the store before parking or is already
+        // registered as a waiter when the notify lands. `offer` discards
+        // events once `closed` is set, so this notify is the last one the
+        // loop can ever receive and the join below depends on it.
+        {
+            let _inbox = self.inner.lock_inbox();
+            self.inner.closed.store(true, Ordering::Release);
+            self.inner.available.notify_one();
+        }
         if let Some(handle) = self.thread.take()
             && handle.thread().id() != std::thread::current().id()
         {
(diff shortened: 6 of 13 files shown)


mtopolnik:
## Review

**Verdict: approve with comments.** The fix is correct and nothing here blocks merging. There are two one-line changes to the new test I'd suggest making first, plus a few smaller notes.

---

### The bug is real and the fix works

I didn't take the description on trust — I pulled the shutdown handshake out into a standalone program and ran the same create/drop cycle against both versions of `Drop`:

- **Before the fix:** hung on all three runs, after 1181 / 10729 / 356 cycles.
- **After the fix:** finished all 20,000 cycles, three runs out of three.

Someone also reverted just the `Drop` body in a working copy and ran the new test against it: it hangs until the 60-second timeout without the fix, and passes in about a second with it.

The reason the fix works is that `dispatch_loop` keeps holding the inbox mutex from the moment it reads `closed` right through to parking on the condvar — `Condvar::wait` signs the thread up as a waiter before it lets go of the mutex. Before the fix, `drop` set the flag and signalled without taking that mutex, so the signal could land in the gap when nobody was listening yet. Since `offer` throws events away once `closed` is set, no later signal could arrive, and the `join` waited forever. Doing both steps under the same mutex leaves only two possible orderings, and both end with the thread exiting.

I walked through the other situations the thread could be in when `drop` runs — freshly spawned but not yet running, already parked, part-way through a listener callback, never started because `Builder::spawn` failed, and holding a poisoned mutex. All of them end correctly. The new critical section can't panic either (`lock_inbox` recovers from poisoning rather than unwrapping), which matters because the FFI crate builds with `panic = "abort"`.

**No new deadlock.** The inbox mutex is a leaf lock — only `offer`, `dispatch_loop`, and now `drop` ever take it, each does a tiny amount of work, none of them takes another lock while holding it, and none of them runs user code. `dispatch_loop` lets go of the guard before calling the listener, so a listener that closes the pool from inside the callback still gets the lock immediately, and the self-join check keeps it from joining itself. I also checked that no pool lock is being held when `QuestDb::drop` reaches `conn_events.close()` / `rejections.close()` — each guard in that function is scoped to its own block.

**The "only instance" claim in the description checks out**, and is actually stronger than stated: there are only five condvars in the whole crate, and the only two plain unbounded waits are this one and `BackpressureNotifier::wait_for_change`, which is protected by its generation counter under its own mutex. Everything in `db.rs` uses `wait_timeout`, so a missed signal there would cost a little latency rather than hang.

---

### Worth changing

#### 1. The new test does about ten times more work than it needs to

Running the test against the old code with a round counter added, it caught the hang on 12 out of 12 attempts, and the latest it ever took was **round 394 of 20,000**. That works out to roughly a 1-in-219 chance per round, so 2,000 rounds already gives better than 99.99% detection. Twenty thousand doesn't buy meaningfully more confidence — it just costs more time.

That cost gets multiplied. `conn_events` is compiled whenever `_sender-qwp-ws` is on, which covers 6 of the 9 `questdb-rs` cargo invocations in `ci/run_all_tests.py`, across Linux, macOS, windows-2022 and windows-2025 — roughly 120,000 thread create/joins per CI leg and about a million per pipeline run.

Here's how it scales with machine load on my box (10 cores, debug build):

| Machine state | Test time | Result |
|---|---|---|
| Idle | 0.46–0.57 s | 5/5 pass |
| 3× oversubscribed | 1.3–3.0 s | 20/20 pass |
| 10× oversubscribed | 7.2–9.6 s | 8/8 pass |

It degrades smoothly with load and never stalls, so there's no hidden problem — but two people reviewing this did hit the 60-second timeout on a machine that was busy compiling, and Windows thread creation is several times more expensive than macOS (bigger stacks, CRT setup, and security software hooking thread creation). Windows is the one platform nobody has timed.

Dropping `20_000` to `5_000` costs nothing in detection power and cuts the worst case by 4×.

#### 2. When the test does fail, you can't tell why

The message is `"a dispatcher drop stalled joining its thread"`, with no indication of how far it got. A genuinely stalled join and a starved CI agent look identical, which means the natural reaction to a red build is "flaky, re-run" — exactly the reaction that would let this bug back in.

Adding the round count separates the two cleanly, since a real stall shows up in the low hundreds and a slow machine shows up in the tens of thousands:

```rust
let rounds = Arc::new(AtomicU64::new(0));
// in the worker: rounds_in_thread.store(round as u64, Ordering::Relaxed);
match done_rx.recv_timeout(Duration::from_secs(60)) {
    Ok(()) => {}
    Err(RecvTimeoutError::Timeout) => panic!(
        "only {} of 20000 drop/join cycles finished in 60s — either a stalled \
         join or a badly overloaded machine",
        rounds.load(Ordering::Relaxed)),
    Err(RecvTimeoutError::Disconnected) => panic!("the drop/join worker panicked"),
}
```

Splitting `Timeout` from `Disconnected` also matters: right now, if the worker thread panics for any reason, the channel closes, `recv_timeout` returns an error immediately, and the test fails instantly while claiming a join stalled.

#### 3. The test can quietly stop testing anything

`EventDispatcher::named` swallows a thread-spawn failure: it logs a warning, sets `closed`, and leaves `thread` as `None`. If spawning ever fails, `drop` takes the lock, sets a flag nobody reads, skips the join, and the round proves nothing. Twenty thousand rapid create/join cycles is exactly the workload that could bump into a thread limit on a constrained agent, and the loop would then race through and report success. The warning wouldn't be visible either, since no logger is set up in these tests.

The test is in the same module, so `assert!(dispatcher.thread.is_some())` inside the loop closes this off in one line.

#### 4. One sentence in the new comment claims more than `offer` delivers

> `offer` discards events once `closed` is set, so this notify is the last one the loop can ever receive and the join below depends on it.

`offer` checks `closed` and signals *outside* the inbox mutex, so on its own it doesn't guarantee that. What actually guarantees it is that both users of the dispatcher — `ConnectionEventSource::offer` and `RejectionEventSource::publish` — hold their own `Mutex<Option<EventDispatcher<_>>>` across the whole `offer` call, while `close()` needs that same mutex to take the dispatcher out. That's the property a future third user would need to preserve, and since `EventDispatcher` is publicly constructible (`questdb::ingress::conn_events`, with a `pub fn new`), someone could write one against this comment and skip it.

The second half is loose too: the join depends on the loop re-reading `closed` under the mutex before each park, not on this being the final signal. Extra signals are harmless.

The first sentence — the actual explanation for the fix — is accurate.

#### 5. The description overstates who's affected

> Every `questdb_db_connect_with_handlers` installs a rejection handler, so **every pool spawns this thread**

Not quite. A dispatcher is only built when a listener or error handler is actually supplied; `QuestDb::connect()` passes the default (both `None`) and gets `ConnectionEventSource::disabled()` and `RejectionEventSource::logging_default()`, neither of which starts a thread. Both callbacks are nullable at the FFI layer, and `client.h` documents a NULL rejection callback as selecting the logging default.

So the affected group is pools created *with* a listener or rejection handler. That does include py-questdb-client, so the reported hang stands — the claim is just wider than the code supports.

#### 6. The test's use of a background thread is deliberate but unexplained

I confirmed that when the test fails, the process still exits properly (exit code 101, other tests in the module still report ok). That works because libtest exits the process without joining background threads, so the stuck worker and its parked dispatcher just get killed.

But that only holds because the loop runs on a `std::thread::spawn` and the test body waits on a channel with a deadline. Written inline, libtest would block forever joining its own test thread and hang the entire `cargo test` with no output at all — which is the production failure mode. The comment explains the race but not this constraint, so someone could "simplify" the spawn away later. One sentence would prevent that.

---

### Test coverage

**One gap worth closing (not blocking).** No test closes a `RejectionEventSource` that never received anything — which is the exact path in the reported crash (`questdb_db_close → RejectionEventSource::close → EventDispatcher::shutdown → pthread_join`). Both existing rejection tests publish an event and wait for it to be delivered before closing, so the dispatcher has been parked for a long time by then, and `logging_default_counts_deliveries` has no dispatcher at all. No pool or FFI test closes a pool holding an error handler with zero rejections either.

This is softened by `Drop for EventDispatcher<T>` being a single generic implementation, so the new `EventDispatcher<ConnectionEvent>` test does exercise the same code — only the concrete `T` differs. Still, two lines (build a `with_handler` source, close it without publishing, loop a few hundred times) would give the reported path its own test rather than having it inherit one.

**One caveat to record.** This is a race-window test, not a proof. On a genuine single-core agent, the parent would finish store → notify → join before the child ever got scheduled, so the child would see `closed` already set and the bug would be unreachable — the test would pass against the old code. Every hosted agent has at least 2 vCPUs, so this is theoretical.

**Worth knowing:** no sanitizer, miri, or emulated-architecture job runs the Rust suite — `SanitizeCppTests` runs `ctest -E rust_tests`, which excludes it. ThreadSanitizer would catch this class of bug directly and isn't running, so this loop is the only detector the project has for missed wakeups. That's a good argument for keeping it in the normal test suite rather than marking it `#[ignore]`, and for making its failure message trustworthy. Leaving it un-ignored also matches how the repo already works: `#[ignore]` here marks benchmarks, tests needing `--test-threads=1`, tests needing external resources, and fixture regenerators — not slow tests. Several 10,000-iteration loops already run normally.

---

### Things I checked and ruled out

- **"The hang is really in the failing connect, not in close."** The sampled stack in the description shows otherwise: `questdb_db_close → QuestDb::drop → RejectionEventSource::close → EventDispatcher::shutdown → pthread_join`.
- **"`offer` signals outside the lock too, so it has the same bug."** It doesn't. What `offer` changes is "is the inbox non-empty", and it changes that *inside* the mutex. The waiter re-checks that under the lock, so signalling outside is fine. The original bug was specific to `closed`, which was changed under no mutex at all.
- **"The store, not the signal, is what needs to be inside the mutex."** Either placement works, because the mutex serialises against the loop's uninterrupted read-then-park section either way. `QuestDb::drop` (store outside, signal under) and this PR (both under) are both correct — the two comments aren't contradicting each other.
- **"`drop` should use a time-limited join like `SyncQwpWsRunner::drop` does."** That would be unsafe here. A join that gives up leaves a thread running that can still call the C callback with `user_data` the caller has already freed, which breaks the promise in `client.h`. The other half of that idea — a `wait_timeout` in `dispatch_loop` — would be safe and would turn any future variant of this bug into a latency blip instead of a hang.
- **"Taking a mutex in `drop` adds contention."** Dispatchers are created once per pool and dropped once. `drop` takes this mutex exactly once in the dispatcher's lifetime, and never during the join.

---

### Noticed nearby — pre-existing, not asking for changes here

These all behave the same before and after this PR. Mentioning them only because they sit on the code path it touches.

- **`catch_unwind` does nothing in shipped builds.** `questdb-rs-ffi/Cargo.toml` sets `panic = "abort"` for both release and dev, so the guard around the listener call is compiled away, and the doc comment saying "Panics are caught and logged; the dispatcher keeps running" isn't true for any FFI or Python consumer — a panicking listener takes the host process down. The existing panic test only passes because cargo ignores the `panic` setting for the test profile.
- **`ConnectionEventSource::close()` throws its counters away**, so `dropped()` / `delivered()` return 0 afterwards, while `RejectionEventSource::close()` deliberately preserves its totals. An FFI-owned sender that outlives its pool reports zero connection events dropped, while the rejection equivalent on the same object still reports properly.
- **Dispatcher threads start before the config is validated.** `connect_with_handlers` builds both sources, and only then does `connect_impl` call `conf::parse(conf)?`. So every connect with a bad config starts up to two threads and drops them microseconds later — the widest possible version of the window, on a path that never needed them. After this fix that's just wasted work; parsing first would remove it entirely.
- **The Rust doc on `EventDispatcher` states the "listener never runs again after drop returns" promise unconditionally**, but it isn't true when drop happens on the dispatcher thread itself — the handle is dropped without joining, so the in-flight callback finishes afterwards. The C header already documents this exception correctly; only the Rust side overstates it.

---

### Checks I ran

- `questdb-rs` library tests: 1690 passed, 0 failed, 23 ignored (34.8 s)
- `questdb-rs-ffi` tests: 88 passed, 0 failed
- Both match the numbers in the description exactly.
- `cargo fmt --check` clean; `cargo clippy --tests` clean
- New test on its own: 5 runs, 0.46–0.57 s each, all passing
- No submodule pointer changed in this PR

One last thing in the fix's favour that's easy to miss: it also removes latent hangs from the existing tests. Before it, any pool or FFI test that closed a dispatcher which had never received an event could block in `pthread_join` inside libtest and hang the whole `cargo test` run with no output whatsoever.

---
<sub>Review produced with [Claude Code](https://claude.com/claude-code).</sub>

mtopolnik:
Split: the GEOHASH precision change that had been sitting on top of this branch (`79e3ebb9`) moved to #193. This PR was force-pushed `79e3ebb9` → `006767e2`, dropping only that tip commit — the four dispatcher commits and their review history are unchanged, and the title and body already described only the dispatcher fix, so they now match the contents.

The two halves touch disjoint files (`conn_events.rs` / `rejection_events.rs` / `questdb` here; the three `column_sender/` encoders in #193), so each is independently reviewable and mergeable in either order. Each half's diff is byte-identical to the corresponding subset of the original PR — nothing was rewritten, dropped, or duplicated.

Verified on this branch after the split: `cargo fmt --check` clean, and the 12 `conn_events` + `rejection_events` tests pass.

mtopolnik:
## Review — level 3

Base `8ef0219f` → head `006767e2`, +188/−3.

## The fix is correct

I proved the bug and the fix by reverting *only* the five-line `Drop` hunk and running the real test binary (repeated independently four times, on separate builds):

| | unfixed `Drop` | HEAD |
|---|---|---|
| `drop_joins_…never_saw_an_event` | fails 12/12, 8/8, 6/6 (stalls at rounds 152–1058 of 5000) | passes 10/10, 5/5 |
| standalone protocol model | stalls after 244–764 rounds, 8/8 runs | 5000/5000 rounds, every run |

`dispatch_loop` holds `inbox` from its read of `closed` (`conn_events.rs:334`) through `Condvar::wait`'s atomic enqueue-and-release (`:345`). Publishing the close inside that same mutex leaves exactly two orderings, both terminating. All eight worker states — unspawned, pre-lock, about-to-park, parked, in-listener, post-listener, spawn-failed, self-drop — terminate. The guard is correctly released at `:315` **before** `join()` at `:319`; holding it across the join would deadlock unconditionally.

The "only instance" claim holds: of 47 condvar references, only two are unbounded waits, and `BackpressureNotifier::wait_for_change` bumps its generation under its own lock.

No lock-order edge is added. Ten callsite groups were walked (pool drop, both sources, sender/QWP plumbing, orphan drainer, FFI entry points, pool tests) and all returned SAFE. Both `close()` bodies are byte-identical to base, and the `MutexGuard` temporary in `let dispatcher = self.lock_dispatcher().take();` releases at the semicolon, so the join never runs under the source mutex. The dispatcher close also sits *after* the bounded QWP drain, so no store-and-forward budget is affected.

---

## Moderate

### 1. The startup handshake cuts the rejection test's detection rate from 12/12 to 3/12

`rejection_events.rs:228-232` — in-diff.

Commit `006767e2` is titled "Prove the dispatcher worker ran before closing it". Controlled ablation, one binary, one machine, 12 runs each, only the guard changed:

| guard in `close_joins_a_handler_dispatcher_that_never_saw_a_rejection` | detected | stall rounds |
|---|---|---|
| `wait_for_dispatcher_start(10s)` (as merged) | **3/12** | 0, 0, 0 |
| plain dispatcher-presence check | **12/12** | 0, 294, 150, 149, 309, 311, 784, 687, 256, 616, 213, 365 |

Mechanism: `signal_worker_start` fires at `conn_events.rs:326`, *before* the worker takes the inbox lock. The closer must then be woken from `wait_timeout`, reacquire `worker_started`, return, and run the `round % 400` spin sweep — by which time the worker is already a registered waiter, past the window. Only round 0 retains any chance, which is why every detection lands there. The spin sweep, which does useful work in the conn-events test, sweeps nothing here.

Independent confirmation on other builds: 3/8 and 5/10. Pooled across all four measurements: 11/42.

The regression itself stays guarded — the conn-events test hits the same `EventDispatcher::drop` 12/12 — so this is not blocking. But the test covering the *reported* crash shape (`RejectionEventSource::close` with zero rejections) is the unreliable one, and the mechanism added to prove non-vacuity is what broke it.

`thread.is_some()` already rules out the vacuous pass it was written for: once spawn succeeds, `Drop` unconditionally joins, and a returning join proves the worker ran. Suggest deleting `worker_started`, `signal_worker_start`, both `wait_for_worker_start`s and `wait_for_dispatcher_start`, and forwarding a `thread.is_some()`-backed `has_dispatcher_thread()` instead. That also removes `conn_events.rs:325-326` — the only `#[cfg(test)]` line inside `dispatch_loop` — so the tests would exercise the production loop rather than a test-shaped variant.

(The ablation above used a weaker presence check purely to isolate the variable; the forwarder is the right form.)

---

## Minor

**2. `rejection_events.rs:246-248` — the triage hint is wrong for this test.** It says "a stalled join reports a few hundred, an overloaded machine reports most of them". Measured stall counts for this test are `0, 0, 0`. The identical string at `conn_events.rs:723-728` does match observation (152–608). More generally the count carries little signal — a genuine stall at 20 000 rounds reported 8892. Sampling `completed` twice ~1 s apart and reporting frozen-vs-advancing would be sound.

**3. `conn_events.rs:303-304` — one sentence of the new comment claims more than the code delivers.** "every offered event either delivered or counted in `dropped`" holds only for events reaching `EventDispatcher::offer`. At source level the existing test `close_fences_handler_and_keeps_counters` (`rejection_events.rs:200-204`) asserts the opposite: publish after close is counted nowhere. Suggest scoping the sentence to "every event that reaches `EventDispatcher::offer`". Paragraph 1 — the actual explanation of the fix — is accurate in every claim.

**4. `rejection_events.rs:126-130` holds the source dispatcher mutex across a ≤10 s blocking wait** (the `MutexGuard` temporary spans the whole `is_some_and`). Test-only and harmless today, since the sole caller is single-threaded. Moot if finding 1 is taken.

**5. The description overstates the blast radius.** "Every `questdb_db_connect_with_handlers` installs a rejection handler, so every pool spawns this thread" is contradicted by `db.rs:796-802` (`None → logging_default()`, no dispatcher, no thread), by `column_sender.rs:1467` (`if let Some(cb) = rejection_callback`), and by the header this PR ships alongside, `client.h:201-206`, which documents NULL as selecting the logging default. Commit `6574fa77`'s body carries a broader version of the same claim. Accurate framing: any pool built with a rejection handler **or** a connection listener — which includes py-questdb-client, so the report stands.

**6. Commit `680ff860 "update questdb module"`** is lowercase where its siblings capitalise, vague, and has an empty body. Separately, the submodule bump is unrelated to the dispatcher fix and changes the server every integration test runs against, so a red CI on this PR is ambiguous. Provenance is clean — worth landing separately.

**7. Public `offer` rustdoc (`conn_events.rs:242-243`) doesn't state the serialization requirement the new `Drop` comment relies on.** `EventDispatcher` is `pub` with `pub fn new`/`offer`; the invariant that makes the accounting hold is supplied by the two in-crate holders and is recorded only in a private comment. Pre-existing, but this PR is what documented it — one sentence on `offer` is cheap and in scope.

**8. The 60 s `recv_timeout` is a 6× outlier** (every other in the crate is 1–10 s). Measured worst case under 4× CPU oversubscription is 0.72 s, a ~90× margin. 10–15 s would keep it safe and cut a genuine-regression CI run from ~6 min of dead waiting to ~1.5 min across six feature legs.

---

## Coverage gaps

No Critical gaps. Three Moderate:

- **Zero-event `ConnectionEventSource::close()`** — `close()` is exercised only by `source_close_fences_listener`, which delivers an event first; the new test drops the raw dispatcher instead. One line in the existing loop.
- **Pool/FFI close with handlers installed and zero events** — nothing at any level. Every `qwp_sender_pool.rs` conn-event test connects first; the one FFI handler test waits for `Connected` before closing; `questdb_db_connect_with_handlers` has no test anywhere. This is the shape that was actually reported. A `connect_with_handlers` → immediate-drop test is the true end-to-end guard.
- **Re-entrant close from inside a listener (self-join guard, `conn_events.rs:316-318`)** — untested on any surface, though documented at `client.h:138-141`. Newly relevant: `Drop` now takes the inbox mutex, and that path is safe only because `dispatch_loop` releases the guard before invoking the listener. Nothing pins that invariant.

Accepted: the spawn-failure path (needs process-global rlimit manipulation; the current guards already stop it producing a false pass).

---

## Checked and dismissed

- **`ConnectionEventSource::close()` zeroes the public counters forever** (no `closed_*` shadow, unlike `RejectionEventSource`) — reachable from C via `line_sender_connection_events_dropped` on an FFI-owned sender outliving its pool. Real, but byte-identical to base. Pre-existing; worth a separate issue.
- **`RejectionEventSource::close()` publishes counters after releasing the mutex**, so a concurrent reader sees 0 for the join's duration — also byte-identical to base.
- **Event pushed after the worker exited, counted nowhere** — unreachable in safe Rust: `offer` takes `&self`, `Drop` takes ownership, and both sources serialise on their own mutex regardless.
- **`T: Drop` re-entering the dispatcher under the inbox guard** — contrived; both concrete `T`s are plain data.
- **"Panics are caught and logged" (`conn_events.rs:120-122`) is false under `panic = "abort"`** — pre-existing, unchanged.
- **`notify_one` under the mutex costs performance** — one futex round-trip once per dispatcher lifetime, <0.1% of a close that also joins the reaper and tears down sockets; matches the existing `db.rs:1435-1447` idiom.
- **New tests are too expensive for CI** — measured ~0.9 s added to a ~35 s lib run, ~45 invocations per pipeline.
- **"Waiting for worker start wouldn't change the race window"** — disproved by ablation; it is worse than neutral.

---

## Summary

**Approve with comments.** No Critical findings, no Critical coverage gaps, nothing blocking merge. The fix is correct, minimally scoped, matches the notify-under-mutex pattern already used in `db.rs`, adds no panic site on any FFI-reachable path, and introduces no lock-order edge.

The one thing worth changing before merge is finding 1: the last commit's stated goal was to make the close/join test prove more, and it measurably made that test prove less. Findings 2, 3 and 5 are one-line corrections.

- **Test gate:** `cargo fmt --check` clean; `cargo clippy --tests` clean on `questdb-rs` across three feature configurations (including the minimal `_sender-qwp-ws` build — zero dead-code warnings) and on `questdb-rs-ffi`; full `--lib` suite 1691 passed / 0 failed / 23 ignored. The one clippy hit is pre-existing in `polars.rs:679`, untouched here.
- **Coverage gaps:** 0 Critical, 3 Moderate.
- **Submodule:** `questdb` `3c1c3382 → 271ec07c` — UPSTREAM-SYNC (contained in `origin/master`); contents out of scope.
- **Candidates:** 8 admitted, 8 omitted as false or not attributable.
- **In-diff / out-of-diff:** 8 in-diff, 0 out-of-diff admitted. Two out-of-diff candidates were found (the counter-shadow asymmetry and the counter-publication window); both were omitted as byte-identical to base.

bluestreak01 (approved):

PR #190, opened 2026-08-22
feat(qwp): non-blocking completed_fsn poll; retain size-hint scratch

nwoolmer:
## Summary

Two related QWP/WebSocket changes.

**1. New public Rust API: `Sender::completed_fsn(AckLevel)`** (`b1d44148`)

A non-blocking poll for the completion watermark — the polling counterpart to
`Sender::wait`. `AckLevel::Ok` reports the highest FSN the server has accepted;
`AckLevel::Durable` reports durable-ACK coverage.

Motivation (questdb/relay): with `request_durable_ack=on`, blocking on `wait`
per batch costs too much latency for an MQTT firehose. The relay polls the
watermark instead and releases pending data without blocking the sender.

Contract, as documented on the method:

- Like `wait`, `AckLevel::Durable` requires QuestDB Enterprise and a sender
  opened with `request_durable_ack=on`; otherwise the call is rejected with
  `InvalidApiCall` up front, ahead of any terminal error the sender holds, so a
  caller polling for durability cannot silently read acceptance coverage.
  `acked_fsn()` is unchanged: it reports the watermark at the sender's
  configured level and is never rejected on that account, so it equals
  `completed_fsn(AckLevel::Durable)` only under the opt-in.
- In background progress mode with durable ACKs, `Ok` advances ahead of
  `Durable`. Manual progress mode has no separate OK tracker, so both levels
  report the completed watermark even under `request_durable_ack=on`.
- The poll makes no progress itself (manual mode needs `drive_once`) and does
  not dispatch buffered rejections to an error handler; `flush`,
  `flush_and_get_fsn`, `wait`, `drive_once` and `close_drain` do.

**2. Perf: retain the size-hint dirty-index allocation** (`9dbca939`, `6410beb8`)

`QwpWsSizeHint::len()` drains `dirty_tables` in place instead of taking the
Vec, so a reusable buffer no longer allocates and frees the index on every
`len()` call that follows an append. The only in-repo caller is
`line_sender_buffer_size` via `Buffer::len`; the saving is one small
malloc/free per size query on a reused buffer.

## Contract change for external callers

`completed_fsn(AckLevel::Durable)` on a sender opened **without**
`request_durable_ack=on` previously returned `Ok(acceptance watermark)`; it now
returns `Err(InvalidApiCall)` on every call. `completed_fsn` is unreleased and
has no in-repo caller outside tests, but questdb/relay consumes it: a relay that
polls `Durable` on a non-opt-in sender must switch to `AckLevel::Ok` or
`acked_fsn()`, or enable the opt-in.

The `InvalidApiCall` message for `AckLevel::Durable` without the opt-in is now
one subject-free string shared by the row-major `Sender`, the column-major pool
backend and the direct column sender:
`AckLevel::Durable requires `request_durable_ack=on` in the connect string.`
This changes the message text (not the code) returned by the released
`Sender::wait` / `line_sender_qwpws_wait` / `line_sender::wait` and the
column-major `wait`; nothing in-tree matches the old text.

## Follow-ups on this branch

- `qwp_ws_completed_fsn` matches `AckLevel` exhaustively, no wildcard arm
  (`a6370454`).
- Docs corrected: manual-mode omission, ordering guarantee scoped to a single
  read pair, error-handler dispatch points, cross-references from
  `flush_and_get_fsn` / `acked_fsn` / the ingress guide (`45a35723`,
  `00bb0540`, `280db7c3`).
- Tests: `wait` rejects non-QWP/WS senders; manual mode with durable ACKs
  reports one watermark after the server OK is consumed; `completed_fsn`
  returns the terminal error on a rejected sender (`b1b4296a`, `f72e72ed`,
  `87c0f40b`).
- Review at `87c0f40b` (request changes): `completed_fsn(Durable)` enforces
  the same opt-in as `wait`, shared with `wait_inner` through one helper
  (`51614908`); publication is asserted ahead of acceptance with both
  completion levels and `acked_fsn` still `None` (`c5d782f9`); a gated
  mock pins that the poll surfaces a terminal reject without dispatching the
  error handler, that the diagnostic survives, and that the next `flush`
  delivers it exactly once (`e3d2e8e8`, `1c55ff30`); `ok_fsn` joins the
  lock-free watermark read test (`23ee78c0`).
- Terminal-state coverage of the `Durable` poll restored on a durable-ACK
  sender, and `acked_fsn` asserted terminal in both progress modes
  (`c5f56854`); opt-in message unified (`0eab85f2`, `2d0b292f`); docs
  tightened (`74f3e5d5`).

## Scope

`completed_fsn` lands on the Rust row-major `Sender` only. The sibling
watermark accessors (`published_fsn`, `acked_fsn`) also exist on
`BorrowedSender`, the column-major sender, and the C and C++ APIs; this is a
deliberate first step, not an oversight. The column-major backend already uses
`qwp_ws_ok_fsn_background` internally, so extending the surface later is
mechanical. C/C++ callers keep the blocking `wait` for now.

## Validation

- full `questdb-rs` lib suite: 1567 passed, 0 failed, 21 ignored
- `cargo fmt` clean; `cargo clippy --tests` introduces no new warnings;
  `cargo doc` and `ci/check_docs.py` clean

## Summary by CodeRabbit

* **New Features**
  * Added non-blocking completion watermark polling for QWP/WebSocket senders at “OK” and “Durable” levels.
  * Manual progress mode reports a unified completed watermark.
  * Terminalized senders now surface status errors through polling.

* **Performance**
  * Improved buffer reuse during repeated size calculations.

* **Documentation**
  * Clarified completion, durable acknowledgment, reconnect recovery, and local durability behavior.
  * Durable completion polling requires durable acknowledgments to be enabled.


eugenels (commented):
Reviewed at `9dbca93`. The perf change is correct, and the new regression test genuinely fails on base — base's `mem::take` leaves `dirty_tables` at capacity 0, so `assert!(capacity >= 1)` trips. No blocking issues.

Three inline suggestions below, plus two coverage gaps and two PR-level notes here. Everything was verified by applying it: all suggestions together on top of `9dbca93` give `cargo fmt --check` clean, no new clippy warnings (the 15 existing ones are all in untouched files), and 7/7 passing on `cargo test --features sync-sender-qwp-ws,sync-sender-http --lib -- qwp_ws_cached_size sender_completed_fsn`.

## Coverage gaps

Both new tests use `request_durable_ack=on` background mode. Nothing covers the manual handler that `completed_fsn` explicitly matches, and nothing covers the non-durable config — so a later change to either behaviour breaks no test. Both harnesses already exist in the file. These compile and pass:

```rust
/// Without `request_durable_ack=on` there is no durable watermark to report.
/// The poll must refuse `Durable` exactly as `wait` does, rather than handing
/// back the plain-OK watermark to a caller asking about durability.
#[test]
fn sender_completed_fsn_rejects_durable_without_durable_ack() {
    let (port, _rx) = spawn_mock_server();
    let conf = format!("ws::addr=127.0.0.1:{port};");
    let mut sender = SenderBuilder::from_conf(conf).unwrap().build().unwrap();

    let mut buf = sender.new_buffer();
    buf.table("trades")
        .unwrap()
        .column_i64("qty", 1)
        .unwrap()
        .at_now()
        .unwrap();
    let fsn = sender.flush_and_get_fsn(&mut buf).unwrap().unwrap();

    assert!(
        wait_until(Duration::from_secs(5), || sender
            .completed_fsn(crate::ingress::AckLevel::Ok)
            .unwrap()
            == Some(fsn)),
        "OK watermark must cover the published frame"
    );

    let err = sender
        .completed_fsn(crate::ingress::AckLevel::Durable)
        .expect_err("Durable poll must be rejected without request_durable_ack=on");
    assert_eq!(err.code(), ErrorCode::InvalidApiCall);
    assert!(
        err.msg().contains("request_durable_ack=on"),
        "error should name the missing setting, got: {}",
        err.msg()
    );
    // Same rejection as the blocking sibling, so migrating from `wait` to the
    // poll cannot silently drop the guardrail.
    let wait_err = sender
        .wait(
            crate::ingress::AckLevel::Durable,
            Duration::from_millis(100),
        )
        .expect_err("wait must reject Durable without request_durable_ack=on");
    assert_eq!(wait_err.code(), err.code());
}

/// Manual progress mode has no separate OK tracker: `qwp_ws_ok_fsn_manual` and
/// `qwp_ws_acked_fsn_manual` both read the completed watermark. Pin that, so
/// the documented "Ok advances ahead of Durable" split is not read as a
/// promise that holds in this mode too.
#[test]
fn sender_completed_fsn_manual_mode_levels_coincide() {
    let (port, _rx) = spawn_mock_server();
    let conf = format!("ws::addr=127.0.0.1:{port};qwp_ws_progress=manual;");
    let mut sender = SenderBuilder::from_conf(conf).unwrap().build().unwrap();
    assert_eq!(
        sender.completed_fsn(crate::ingress::AckLevel::Ok).unwrap(),
        None
    );

    let mut buf = sender.new_buffer();
    buf.table("trades")
        .unwrap()
        .column_i64("qty", 1)
        .unwrap()
        .at_now()
        .unwrap();
    let fsn = sender.flush_and_get_fsn(&mut buf).unwrap().unwrap();

    assert!(
        wait_until(Duration::from_secs(5), || {
            let _ = sender.drive_once();
            sender.completed_fsn(crate::ingress::AckLevel::Ok).unwrap() == Some(fsn)
        }),
        "manual OK watermark must cover the published frame once driven"
    );
    assert_eq!(
        sender.completed_fsn(crate::ingress::AckLevel::Ok).unwrap(),
        sender.acked_fsn().unwrap(),
        "manual mode reports one watermark for both levels"
    );
}
```

The first test assumes the `Durable` guard suggestion is applied; drop its `expect_err` half if you keep the permissive behaviour.

## PR-level

**The title and description don't mention the new public API.** Title is `perf(qwp): retain websocket size-hint scratch` and the Summary describes only the size-hint change, but `b1d44148` adds `pub fn Sender::completed_fsn(AckLevel)`. Only the auto-generated CodeRabbit block mentions it — anyone reading title + Summary (a reviewer, a release-notes author) would miss a new public Rust API entirely. Either split into two PRs or retitle and describe both.

**`completed_fsn` lands on one surface only.** Every sibling watermark accessor exists on five:

| Surface | `published_fsn` | `acked_fsn` | `completed_fsn` |
|---|---|---|---|
| Rust `Sender` | `sender.rs:641` | `sender.rs:661` | **added** |
| Rust `BorrowedSender` | `db.rs:1789` | `db.rs:1800` | — |
| Rust column-major | `column_sender/sender.rs:834` | `:841` | — |
| C | `line_sender.h:2071`, `qwp_sender.h:1509` | `line_sender.h:2081`, `qwp_sender.h:1521` | — |
| C++ | `line_sender.hpp:2080`, `qwp_sender.hpp:813` | `line_sender.hpp:2093`, `qwp_sender.hpp:824` | — |

The column-major backend already imports and uses `qwp_ws_ok_fsn_background` (`column_sender/sender.rs:39`, `:420`, `:1975`), so the capability is there and simply unexposed; C/C++ callers get the blocking `wait` only. Fine as a first step — just say so in the PR body so it isn't read as an oversight.

## Minor

`qwp_ws_completed_fsn`'s `_ =>` arm (`sender.rs:786-789`) still says `"wait is only supported for QWP/WebSocket senders."` and is now unreachable from both callers, since each guards the handler variant first. Worth deleting, or making the message name the caller.

`completed_fsn` takes `&self`, so unlike `wait` (`sender.rs:704`) it can't drain buffered server-rejection notifications into the installed `qwp_ws_error_handler` (`sender.rs:307-323`). Terminal errors still surface via `check_error()`, but a pure poll loop with no intervening `flush`/`drive_once`/`close` won't dispatch retriable rejections, and they can be dropped (cf. `sender_errors_dropped_total`). One doc sentence pointing at `poll_qwp_ws_error` would close it.

questdb-rs/src/ingress/buffer/qwp.rs
```diff
@@ -2529,8 +2529,11 @@ impl QwpWsSizeHint {
     }
 
     fn len(&mut self, tables: &[QwpWsTableBuffer]) -> usize {
-        let dirty_tables = std::mem::take(&mut self.dirty_tables);
-        for table_idx in dirty_tables {
+        // Preserve the dirty-index allocation across size queries. Taking and
+        // dropping this vector allocated once per row after every buffer clear,
+        // even though the buffer itself is explicitly reusable.
+        let mut dirty_tables = std::mem::take(&mut self.dirty_tables);
+        for table_idx in dirty_tables.drain(..) {
```
eugenels: The `mem::take` is unnecessary here. `Drain` borrows only the *place* `self.dirty_tables`; everything the loop body touches (`self.tables`, `self.encoded_tables_len`, `self.symbol_dict_*`, `self.recomputed_tables`) is a disjoint field reached through a direct field path in the same function, so the borrow checker splits them. `mem::take` is the workaround for a loop body that needs `&mut self` as a whole — e.g. calling `self.mark_dirty(..)` — and nothing here does.

`drain(..)` retains the allocation on its own, which is the entire point of the change. Verified: compiles, and all three size-hint tests pass including the new `qwp_ws_cached_size_hint_retains_dirty_index_capacity`.

```suggestion
        // Drain in place: `Vec::drain` empties the vector while retaining its
        // allocation, so a reusable buffer stops re-allocating the dirty index
        // once per row after every `clear()`.
        for table_idx in self.dirty_tables.drain(..) {
```

questdb-rs/src/ingress/buffer/qwp.rs
```diff
@@ -2548,6 +2551,8 @@ impl QwpWsSizeHint {
                 self.recomputed_tables += 1;
             }
         }
+        debug_assert!(dirty_tables.is_empty());
+        self.dirty_tables = dirty_tables;
```
eugenels, line 2550: Companion to the suggestion above — both of these go away.

`debug_assert!(dirty_tables.is_empty())` is vacuous: `drain(..)` always empties the vector, including on the `continue` branch, so it can never fire.

`self.dirty_tables = dirty_tables` is an unconditional clobber. Today nothing in the loop pushes, but if someone later adds a `mark_dirty` there, the new indices are silently discarded and the size hint goes permanently stale. Note base was actually safe against that — `for table_idx in dirty_tables` had no reassignment. Draining in place cannot lose entries, and that same future edit becomes a compile error instead.

```suggestion
        }

```

questdb-rs/src/ingress/sender.rs
```diff
@@ -663,6 +663,36 @@ impl Sender {
...
+        if !matches!(
+            &self.handler,
+            SyncProtocolHandler::SyncQwpWs(_) | SyncProtocolHandler::ManualQwpWs(_)
+        ) {
+            return Err(error::fmt!(
+                InvalidApiCall,
+                "completed_fsn is only supported for QWP/WebSocket senders."
+            ));
+        }
+        self.qwp_ws_completed_fsn(ack_level)
```
eugenels: Two problems in this block.

**1. `Durable` drops the guardrail `wait` enforces.** `wait_inner` rejects `AckLevel::Durable` outright when the sender was not opened with `request_durable_ack=on` (sender.rs:719-731). This checks only the handler variant, then dispatches to `qwp_ws_acked_fsn_background` -> `runner.acked_fsn()`, which outside durable mode advances on plain OK acks. So with `ws::addr=127.0.0.1:9000;`:

- `sender.wait(AckLevel::Durable, t)` -> `InvalidApiCall`
- `sender.completed_fsn(AckLevel::Durable)` -> `Ok(Some(fsn))`, carrying no durability guarantee

The doc says **"Prefer this over `wait`"**, so a caller migrating a durable barrier to the poll silently loses the misconfiguration check on the exact input `wait` refuses.

**2. The OK-ahead-of-Durable claim is false in manual progress mode.** It holds only for `SyncQwpWs`. `SyncQwpWsRunner` carries `ok_completed_upper: Arc<AtomicU64>` (qwp_ws.rs:512) which `ok_fsn()` maxes against the completed watermark. `ManualQwpWsHandlerState` (qwp_ws.rs:496-505) has no such field, and the two manual helpers are byte-identical:

```rust
pub(crate) fn qwp_ws_acked_fsn_manual(state: &ManualQwpWsHandlerState) -> crate::Result<Option<u64>> {
    check_manual_driver_error(state)?;
    Ok(state.store.completed_fsn())
}
pub(crate) fn qwp_ws_ok_fsn_manual(state: &ManualQwpWsHandlerState) -> crate::Result<Option<u64>> {
    check_manual_driver_error(state)?;
    Ok(state.store.completed_fsn())
}
```
(qwp_ws.rs:3989-4001)

`request_durable_ack` is a real field on the manual state (qwp_ws.rs:501) and is forwarded into `qwp_extra_headers` (qwp_ws.rs:2677), so `ws::...;qwp_ws_progress=manual;request_durable_ack=on;` is accepted — and there `completed_fsn(Ok)` returns the *durable* watermark. A caller releasing backpressure on `Ok` stalls until durable ACKs land. The behaviour is pre-existing (`wait(Ok)` routes through the same helper); the new public doc asserting the opposite is what this PR adds. This is also exactly the mode the doc's own motivation targets ("a thread that owns the socket and drives progress itself").

If you'd rather keep the permissive `Durable` behaviour, that's defensible — but then say so on the item, because right now the doc reads as if the two calls agree.

```suggestion
    /// In background progress mode with durable ACKs the `Ok` watermark
    /// advances ahead of `Durable`; otherwise the two coincide. In manual
    /// progress mode there is no separate OK tracker, so both levels report
    /// the completed watermark even under `request_durable_ack=on`. `Ok`
    /// never lags `Durable`. Both advance on server ACK or server-side
    /// reject-and-continue, so a rejected frame does not leave the watermark
    /// stuck behind it.
    ///
    /// Like [`Self::wait`], [`AckLevel::Durable`] requires a sender opened
    /// with `request_durable_ack=on` and is otherwise rejected, so a caller
    /// polling for durability cannot silently read the plain-OK watermark
    /// instead.
    ///
    /// Prefer this over [`Self::wait`] wherever the caller must not block —
    /// for instance a thread that owns the socket and drives progress itself,
    /// where a blocking barrier would stall the very transport it services.
    /// QWP/WebSocket only; other protocols return `InvalidApiCall`.
    #[cfg(feature = "sync-sender-qwp-ws")]
    pub fn completed_fsn(&self, ack_level: AckLevel) -> Result<Option<u64>> {
        let request_durable_ack = match &self.handler {
            SyncProtocolHandler::SyncQwpWs(state) => state.request_durable_ack,
            SyncProtocolHandler::ManualQwpWs(state) => state.request_durable_ack,
            _ => {
                return Err(error::fmt!(
                    InvalidApiCall,
                    "completed_fsn is only supported for QWP/WebSocket senders."
                ));
            }
        };
        if ack_level == AckLevel::Durable && !request_durable_ack {
            return Err(error::fmt!(
                InvalidApiCall,
                "completed_fsn with AckLevel::Durable requires the pool to be \
                 opened with `request_durable_ack=on` in the connect string."
            ));
        }
        self.qwp_ws_completed_fsn(ack_level)
```

eugenels:
## review-pr, level 3 — `eaf51ceb..87c0f40b`

Full pass: change-surface map, 10 review agents (correctness, panics/abort, FFI, concurrency, resources, perf, tests, API ergonomics, cross-context callers, fresh-context adversarial), every finding re-verified against source. Build-profile premise: `questdb-rs-ffi` is `panic = "abort"` in dev and release, so any reachable panic aborts the host.

**Verdict:** no correctness, safety, or concurrency defects. All callsites of the changed symbols (`wait_inner`, `acked_fsn`, column-major `wait_for_boundary`, FFI `line_sender_buffer_size` / `line_sender_qwpws_wait`) verified SAFE; the `drain(..)` change is value-identical and adds no panic path. 22 draft findings → 14 verified (5 Moderate, 9 Minor), 8 dropped as false positives.

### Fixed on this branch

| | Finding | Commit |
|---|---|---|
| M1 | PR body described a `Durable` opt-in guard that is not shipped | body rewritten |
| M2 | Manual-mode test's 200×`drive_once` loop raced the mock server's OK; could pass without the server accepting anything | `f72e72ed` — gate on `qwp_ws_totals().acks >= 1` |
| M4 | Doc pointed poll-only callers at `qwp_ws_errors_dropped`, which only counts diagnostics already lost; dispatch happens on `flush`/`wait`/`drive_once`/`close_drain` | `280db7c3` |
| M5 | No test that `completed_fsn` errors on a terminal sender | `87c0f40b` |
| m1 | "reject-and-continue" advance described a path that does not exist (every policy replays or terminalizes) | `280db7c3` |
| m2 | Comment credited `#[non_exhaustive]` for in-crate exhaustiveness (wrong mechanism) | `280db7c3` |
| m3/m4 | `Ok` bullet unqualified for manual mode; ordering phrased as one call; `mod.md` "only one" claim overstated | `280db7c3` |

### Deliberately not changed

- **M3** `AckLevel::Durable` enum doc still says the level "requires" `request_durable_ack=on`; `completed_fsn` accepts it without. Left as is — the permissive poll behaviour is provisional.
- **m5** Perf claim magnitude: no in-repo production path calls `QwpWsColumnarBuffer::len`; the saving is per `len()` call, not per row. Body wording corrected; code fine.
- **m7** `restore_snapshot` (`qwp.rs:2968`) replaces the `Mutex` and discards the retained capacity on every rewind. Pre-existing.
- **m8** `ok_fsn` read order (`completed` before `ok_completed_upper`) is what makes `Ok >= Durable` hold; unmarked. Pre-existing.
- **m9** C header `line_sender_qwpws_acked_fsn` lacks the durable-mode caveat present in `qwp_sender.h`. Pre-existing.
- Out of scope: `capture_snapshot` deep-clones every cell, O(cells) per `set_marker` on a QWP/WS buffer. Worth its own issue.

### Downgraded

`continue` branch in `QwpWsSizeHint::len` stranding a dirty flag (unreachable: hint tables never exceed buffer tables); `-=` underflow (totals equal Σ cached entries at every mutation site); `completed_fsn` absent from C/C++/Python (deliberate, see Scope); race in `sender_completed_fsn_polls_ok_ahead_of_durable` (durable strictly gated on the channel); `Sender: Sync` / FFI reach of the `&self` method (no auto-trait change, not exported).

### External note for the relay

`questdb/relay` polls `completed_fsn(AckLevel::Durable)` in background mode without `request_durable_ack=on`. Under the documented contract that value is server-acceptance coverage, equal to `AckLevel::Ok` at every instant — not durability. If the relay's `release_gate = "durable"` is meant to be durable, it needs the connect-string setting.

bluestreak01 (changes requested):
@nwoolmer — Request changes for the explicit-durability contract: completed_fsn(AckLevel::Durable) must reject missing durable-ACK opt-in rather than report ordinary acceptance as durable completion. Preserve the legacy acked_fsn() behavior. Full verified level-3 review, including non-blocking notes and validation: https://github.com/questdb/c-questdb-client/pull/190#issuecomment-5571545620

bluestreak01:
@nwoolmer — level-3 review of `87c0f40be823484dd0513e625471ac1e4f2fcafb`. **Request changes** for the explicit-durability contract below.

## Critical

**1. Explicit `Durable` polling succeeds without durable coverage — in-diff**  
`questdb-rs/src/ingress/sender.rs:697–707`

Without `request_durable_ack=on`, `completed_fsn(AckLevel::Durable)` returns acceptance coverage:

`completed_fsn(Durable)` → `qwp_ws_acked_fsn_*` → completion advanced by an ordinary OK response.

The added test at `questdb-rs/src/tests/qwp_ws.rs:1989–2027` demonstrates this in both progress modes. In contrast, `wait(Durable)` rejects the same configuration at `sender.rs:758–771`.

A caller replacing a durable barrier with its advertised polling counterpart can release its recovery copy before durability is established. The fallback is documented and intentional, but still makes an explicit durability request succeed with weaker guarantees.

**Fix:** enforce the same durable-opt-in validation as `wait`. Preserve legacy `acked_fsn()` behavior and update the fallback/equivalence documentation and tests.

## Moderate

None.

## Minor

**2. Missing publication-before-acceptance assertion — in-diff**  
`questdb-rs/src/tests/qwp_ws.rs:1889–1899`

The test releases the server’s OK gate immediately after publication. A faulty background `Ok` dispatch returning `published_fsn()` would pass its assertions.

**Fix:** before releasing `ok_tx`, assert publication is `Some(fsn)` while both completion levels remain `None`.

**3. Poll-only diagnostic behavior lacks regression coverage — in-diff**  
`questdb-rs/src/tests/qwp_ws.rs:1477–1526`

The terminal test calls `wait` and consumes the diagnostic before testing completion polling. It does not establish that polling preserves pending diagnostics and leaves callbacks undispatched.

**Fix:** gate a rejection until after publication; observe it through completion polling first, assert no callback delivery, then verify the diagnostic remains available. Optionally include `ok_fsn()` in the existing contention test at `questdb-rs/src/ingress/sender/qwp_ws.rs:5040–5082`.

**4. Commit-subject convention — metadata-only**

Six subjects exceed the requested under-50-character limit, including `b1d44148` and `a6370454`. Shorten preserved subjects or use a compliant squash subject.

- **“The test comment incorrectly says `wait` stops transport progress.”** Dismissed: “stalls the whole pipeline” can validly describe blocked application work (`tests/qwp_ws.rs:1861–1865`). Manual `wait` explicitly drives transport (`sender.rs:795–799`).

## Build profile facts

- Standalone Rust defaults to unwinding; FFI dev/release explicitly abort (`questdb-rs-ffi/Cargo.toml:73–77`). `catch_unwind` cannot recover there.
- Overflow checks use Cargo defaults: enabled in dev/test, disabled in release. Package overrides only disable dependency debug information (`questdb-rs/Cargo.toml:430–434`; FFI manifest:79–83).
- No production allocator override; the test allocator delegates to `System` (`questdb-rs/src/lib.rs:103–156`). LTO, codegen units and stripping are unconfigured.

## Summary

**Request changes** at `87c0f40b` for the explicit-durability contract.

- Four findings confirmed; one false positive removed.
- Three in-diff findings, zero out-of-diff regressions after the wider pass, one metadata finding.
- Validation: **1,566 Rust tests passed**, 21 ignored; six targeted FFI tests passed. Formatting, default Clippy, Rustdoc and checked feature combinations passed.
- C++/Python suites were reviewed statically, not executed; the external Enterprise harness remains unverified.
- No source files changed.

eugenels:
## review-pr, level 3 — `87c0f40b..1c55ff30`

Two full passes (change-surface map, 10 review agents each, per-finding verification): one over the fixes for the `87c0f40b` review, one over the resulting commits. Nine commits pushed.

### The `87c0f40b` findings

| # | Finding | Resolution |
|---|---|---|
| 1 | Explicit `Durable` poll succeeds without durable coverage | `completed_fsn` runs the same opt-in check as `wait`, through one helper shared with `wait_inner`, ordered ahead of any terminal error the sender holds. Pinned in both progress modes for the poll and the barrier. `acked_fsn` unchanged and documented as the configured-level accessor. `request_durable_ack` is a write-once copy of the connect string; the handshake echo check rejects a declining server on every connect path, so the flag can't be `true` without server-side durable ACKs. (`51614908`) |
| 2 | Missing publication-before-acceptance assertion | Before the mock releases the server OK, `published_fsn()==Some(fsn)` while `completed_fsn(Ok)`, `completed_fsn(Durable)` and `acked_fsn()` are all `None`. A dispatch arm wired to the publication watermark fails there. (`c5d782f9`) |
| 3 | Poll-only diagnostic behaviour lacks regression coverage | A gated mock holds the rejection until after publication. The poll surfaces `ServerRejection` with the error handler undispatched — deterministic, since the handler lives only on `Sender` and is invoked only from `drain_qwp_ws_error_notifications` on the caller thread; the pool-only `rejection_sink` is `None` here. `poll_qwp_ws_error` still yields the diagnostic, the next `flush` delivers the notification once, the following `wait` fails without a second delivery. `ok_fsn()` added to the lock-free watermark read test. (`e3d2e8e8`, `1c55ff30`, `23ee78c0`) |
| 4 | Commit subjects over 50 chars | The nine new subjects are 42–49 chars. The six earlier ones are untouched; a compliant squash subject at merge (`feat(qwp): non-blocking completed_fsn poll`) rather than a history rewrite. |

### Found in the second pass and addressed

- The `InvalidApiCall` message said "requires the pool to be opened with" on the row-major `Sender` (no pool) and, via `column_sender/conn.rs`, on the direct column sender and its C surface (also no pool). One subject-free string now serves all three sites; `doc/CONSIDERATIONS.md` names both the sender and the pool. This changes the message **text** (not the code) of the released `Sender::wait` / `line_sender_qwpws_wait` / `line_sender::wait`; nothing in-tree matches the old text. (`0eab85f2`, `2d0b292f`)
- Rewriting the schema-reject test for the new ordering dropped the terminal-state coverage of `completed_fsn(Durable)` that `87c0f40b` added. A durable-upgrade variant of the gated mock now pins that the `Durable` poll, `acked_fsn` and the `Ok` poll all surface a terminal reject on a durable-ACK sender, and the schema-reject test asserts `acked_fsn` terminal in both modes. The stalled-frame and gated-reject mocks collapsed into one `spawn_gated_server(upgrade, Option<reject>)`. (`c5f56854`)
- Docs: "otherwise the two coincide" only ever described manual mode with the opt-in; the Enterprise qualifier and the check-before-terminal ordering are stated on `completed_fsn` as on `wait`; `acked_fsn` is described by the level it reports. (`74f3e5d5`)

### Verified and dismissed (for the record)

Handler dispatch from the runner thread making the negative assertion racy — no path exists for builder-built senders. Mock close racing the reject frame into EOF-first — the frame reader parses buffered input before a new read, and the runner stops on `Terminal`. `Sender` drop blocking or double-panicking on the test channels — no `Drop` impl on `Sender`; the runner thread has already exited. `wait(Durable)` on a terminal manual sender driving the transport — it returns at the opt-in check before `published_fsn()`.

### Contract change for the external consumer

`completed_fsn(AckLevel::Durable)` on a sender opened **without** `request_durable_ack=on` used to return `Ok(acceptance watermark)`; it now returns `Err(InvalidApiCall)` on every call, and on a terminal sender that error precedes `ServerRejection`. `completed_fsn` is unreleased with no in-repo caller outside tests, but questdb/relay polls it: a relay polling `Durable` on a non-opt-in sender must switch to `AckLevel::Ok` / `acked_fsn()` or enable the opt-in. PR body updated accordingly.

### Validation

`cargo test --lib` 1567 passed, 0 failed, 21 ignored at `1c55ff30`; `cargo fmt` clean; default clippy adds no warnings; `cargo doc` and `ci/check_docs.py` clean. `Sender::completed_fsn` has no in-tree caller outside `src/tests/qwp_ws.rs` (FFI, C/C++ sidecars, Python harness, soak, enterprise_e2e, examples and benches checked).

eugenels:
## review-pr, level 3 — `8b190bec`

Full pass over the current head (change-surface map, 10 review agents, per-finding verification). **Approve with comments.** No functional defect; three Minor doc/hygiene items and one coverage gap, all addressed in four commits pushed on top (`2597b984..3475881d`).

### Findings and resolution

| # | Finding | Resolution |
|---|---|---|
| 1 | `completed_fsn` doc called the `Ok` watermark "the highest FSN the server has accepted". It is the completion watermark at the OK level: a cumulative OK completes through a frame the server rejected and continued past, as the `acked_fsn` doc already says. The dispatch list also omitted the `flush_and_keep` variants. | Wording mirrors `acked_fsn`, points at `poll_qwp_ws_error`, lists both `flush_and_keep*` methods. The stale `qwp_ws_ok_fsn_manual` comment ("acceptance and completion coincide", false under `request_durable_ack=on`) corrected. (`2597b984`) |
| 2 | The `_ =>` arms in `check_durable_ack_opt_in` and `qwp_ws_completed_fsn` are unreachable (both callers reject non-QWP/WS handlers first) and each carried its own untested `InvalidApiCall` text; base used `unreachable!`. | Both arms now use the same `unreachable!` the wait loop already relies on. (`7b7125f2`) |
| 3 | Coverage: no test observed `completed_fsn(Ok)` across a durable-mode reconnect. The tracker is a shared, only-raised atomic re-cloned into each reconnected core, so no defect, but the new public contract was unpinned there. | The deep durable backlog test now polls both levels at each checkpoint: `Ok` covers the backlog after the initial OKs and after replay while `Durable` stays `None`; both reach the last FSN once the durable ACK lands. (`75f20394`) |
| 4 | Accepted gap: the unified opt-in message was asserted exactly only on the row-major side; the pool test checked the code alone. | Pool arrow test asserts the same string. (`3475881d`) |

### Verified and dismissed

- Gated mock closes after a fixed sleep and could RST the reject frame: keepalive pings need a pending durable ACK, which only exists after a server OK; the durable-reject mock never sends one. 20 repeated runs of every new test passed.
- Row-major C/C++ `acked_fsn` docs lack the durable caveat: pre-existing, headers untouched here.
- Manual-mode headline bullet: already qualified with "(background progress mode; see below for manual mode)".
- No C/FFI counterpart, `completed_fsn` naming, `Sender: !Sync`, `u64::MAX` saturation, `line_sender_buffer_size` NULL contract, atomic orderings: pre-existing or deliberate scope, no user impact.

### Validation at `3475881d`

`cargo test --lib` 1699 passed, 0 failed, 23 ignored; new/extended tests run 20× with no failure; mutation runs in a scratch copy show 7 of the 9 new/extended tests fail on the mutation they guard. Counting-allocator bench against base confirms the size-hint change removes one malloc and one free per `len()` after an append. `cargo fmt` clean, default clippy adds nothing, `cargo doc` and `ci/check_docs.py` clean. C++ and Python suites reviewed statically. No submodule pointer change.

bluestreak01 (approved):
Reviewing PR #190 at level 3, head 3475881dacbfbcc8a20fcc33bce83749431dd9b2, base 621a6baffdc1611474810e4c9e04cc575722f874.

### Minor

**Problem:** Test documentation incorrectly describes the private fallback.  
**Net impact:** Maintainers receive inaccurate guidance about an internal invariant.  
**Evidence:** At `3475881dacbfbcc8a20fcc33bce83749431dd9b2`, `questdb-rs/src/tests/qwp_ws.rs:2766–2769` contradicts `questdb-rs/src/ingress/sender.rs:848`.

The new comment says the helper returns “a plain error rather than a panic,” but its fallback now uses `unreachable!`. The outer protocol guard still returns an error. Remove the obsolete fallback clause. This comment was absent at base.

### Coverage map

**Test gate: passes. Admitted coverage gaps: 0.**

### Summary

**Verdict: approve.**

- Findings: **0 Critical, 0 Moderate, 1 Minor**; one in-diff, no out-of-diff-breakage.
- Validation: 2,147 core tests passed; targeted Arrow, FFI and WS-only tests passed. Formatting, Clippy, Rustdoc, documentation checks and FFI abort-profile build passed.
- Scratch mutation tests detected reverted capacity retention and incorrect watermark dispatch.
- Panic strategy: standalone core defaults to unwind; FFI dev/release abort (`questdb-rs-ffi/Cargo.toml:73–77`). Overflow checks are enabled in dev/test, disabled in release.
- Submodule provenance: no pointer changes.
- Exact-head [TestVsQuestDBMaster](https://dev.azure.com/questdb/questdb/_build/results?buildId=268268) and Enterprise e2e statuses report success. The new Rust poll is exercised directly by mock tests; C++/Python live suites were not run locally.
- Scope remains Rust-only; manual durable-opt-in mode retains one completion watermark.
- Primary checkout unchanged.

PR #191, opened 2026-08-25
perf(qwp): record lengths for a bookmark instead of cloning the buffer

nwoolmer:
`Buffer::bookmark()` on a QWP/WebSocket buffer captured its rewind point by cloning `tables` and `table_lookup` — everything already buffered. `rewind_to_bookmark` only ever *truncates* back to that point, so the copy is never needed, and its cost is proportional to how full the buffer already is.

A caller that bookmarks once per appended row therefore pays a copy that grows as the buffer fills: quadratic total work for a linear amount of appending.

## The cheap mechanism already exists here

Row-level rollback in the same file solves exactly this problem by recording lengths:

```rust
struct QwpWsRowRollbackMark {
    tables_len: usize,
    current_table_idx: Option<usize>,
    state: BufferState,
    table_mark: QwpWsTableRollbackMark,   // row_count, columns_len, cursors
}
```

and `QwpWsTableBuffer::restore` truncates every column's data back to `row_count`. A bookmark needs the same thing with a mark per table rather than only the current one.

So `QwpWsSnapshot` becomes `tables_len` plus a `Vec<QwpWsTableRollbackMark>`, and the rewind drops the tables added after the mark, restores each surviving table, rebuilds the table lookup only when tables were dropped, and marks dirty in the size hint only the tables it actually unwound.

## Multi-row rewinds (found in review)

The first commit's `restore` delegated to the row-level `rollback_row`, which pops a single cell per column and bails unless `last_written_row` is the row being rolled back. That holds for `rollback_current_row`, where exactly one uncommitted row exists, but a marker or bookmark spans any number of rows: once two or more had been appended after the mark, nothing was popped while `row_count` was assigned back anyway, so the encoded frame carried the discarded cells and symbols.

The fix keeps the per-kind single-cell pop but changes its predicate to "the tail cell belongs to row `from` or later", and the column loops until nothing pops. There is no early-return guard left to be wrong for the multi-row case; row-level rollback is the same loop with `from` equal to the open row. A popped symbol that was new to the column removes only its own lookup entry (its bytes are still in the arena, so it is rehashed and removed directly) instead of sweeping the whole lookup, the encoded-size counters are discounted by one pass over the discarded cells instead of resummed over the survivors, and the decimal scale only unpins once no non-null cell remains (it is pinned by the first non-null value, so this is the exact pre-append state).

## Measured

Per-row cost of bookmark-per-row on a QWP/WS buffer, debug build, taken on the first commit. The fix does not touch the capture path these numbers are dominated by; restore now does work proportional to the cells discarded, where the first commit did none for multi-row rewinds.

| | 4,000 rows | 32,000 rows | ratio |
|---|---:|---:|---:|
| before | 2,094 ns | 3,631 ns | 1.73 |
| after | 664 ns | 674 ns | 1.02 |

Flat afterwards, and ~3× cheaper even at the smaller size.

Found from downstream, where a JSON→QWP ingest path bookmarks once per appended row so an overflowing row can start a fresh frame. In a release build there:

| frame size | before | after |
|---:|---:|---:|
| 100 rows | 517 ns/row | 205 ns/row |
| 2,000 | 988 | 204 |
| 10,000 | 3,083 | 207 |
| 50,000 | 13,155 | 208 |

End-to-end ingest throughput rose 24–54% depending on decode path, with ~39% less CPU in the affected worker threads.

## Tests

Added, all comparing encoded bytes and the cached size hint against a buffer that only ever held the surviving rows, for both `rewind_to_marker` and `rewind_to_bookmark`:

- `qwp_ws_columnar_rewind_unwinds_every_row_after_the_mark` — three rows per table after the mark, a column and a table that did not exist at the mark, new symbol values; fresh connection dictionary so a leaked symbol shows up as a byte difference.
- `qwp_ws_columnar_rewind_matches_a_buffer_without_the_discarded_rows` — sweeps the column kinds across mark positions and discard counts, then appends past the rewind, because fixed-width kinds encode by row index and a leftover cell only shows once a new row lands on its index.
- `qwp_ws_columnar_rewind_discards_an_open_row` — rewind while a row is open, in a table the mark knew and in one created after it.
- `qwp_ws_columnar_rewind_recomputes_only_the_tables_it_unwound` — the narrowed size-hint invalidation.
- `qwp_ws_columnar_rewind_keeps_decimal_scale_pinned_by_surviving_rows` — scale unpinned when the pinning value is discarded, kept when it survives.
- `test_bookmark_rewinds_every_table_it_spans` and `test_rewind_discards_every_row_after_the_rewind_point` through the public `Buffer`; both gated on `_sender-qwp-ws` so the TCP-only, HTTP-only and reader-only CI legs still compile.
- `qwp_ws_columnar_rewind_restores_the_precision_a_discarded_row_repinned` and `qwp_ws_columnar_failed_row_restores_the_precision_it_repinned` — the geohash precision restore, through a rewind and through a mid-row error rollback; the sweep also runs every case a second time from a cleared buffer, which is where the retained precision lives.
- `qwp_ws_rewind_after_a_failed_publication_republishes_only_surviving_rows` — on a live connection: publish and keep, bookmark, let an oversized flush fail (which rolls the connection dictionary back), rewind, republish. The second frame must resume at the dictionary watermark and carry only the symbol the post-rewind row introduced.
- `borrowed_sender publishes only the rows a bookmark rewind kept` (C++) — the pooled buffer is the supported C/C++ route to a QWP/WS row buffer; rewind past three rows and flush to the mock.

The existing bookmark and marker tests pass unchanged.

Rewinding across `Sender::flush` or `flush_and_get_fsn` is not possible: both call `buf.clear()` on success, which drops the snapshot, so a later `rewind_to_marker` reports no marker set and a bookmark handle is rejected as stale. A failed flush, or `flush_and_keep` whether it succeeds or not, leaves the buffer untouched (`encode_ws_replay_message` takes `&self`), so `tables_len` still indexes the same tables.

No timing assertion is included. The effect is a ratio on hardware CI does not control, and a bound tight enough to catch a 1.73× growth would be flaky. The numbers above are evidence for review rather than something to enforce in the suite.

## Geohash precision (found in review)

A geohash column keeps its pinned precision across `clear()` and across a flush, so a reused column that no row writes still encodes the precision the server's column has; the first value of the next batch repins it. Recording lengths left that repin standing: rewinding away the row that set it emptied the column but kept the new precision, so the surviving all-null column advertised a precision no row asked for — and against an existing `GEOHASH(25b)` column the server rejects the batch. The clone-based rewind restored the old value, so this was a regression.

Fixed by keeping the displaced precision on the column and putting it back when a rollback empties the column again. One value is enough: a repin can only happen while the column is empty, and a rollback can only empty a column that was empty at the mark (marks are taken on row boundaries), so the pair is exact. Nothing is added to the capture path, and the column enum's size is unchanged.

## Docs

The `clear()` docs in `line_sender.h`, `line_sender.hpp` and `questdb-rs-ffi/src/lib.rs` now state that clearing discards any stored marker or bookmark.

## Unrelated test fixes

Not part of the rewind change, but on this branch:

- `tests/qwp_ws.rs`, stalled-orphan mocks: `qwp_ws_background_orphan_close_is_bounded_and_leaves_orphan_recoverable` failed on windows-msvc-2025 because the mock's 6s park expired before the test sent its cleanup release, so `send(()).unwrap()` hit a disconnected channel. The parks are now 10s and the cleanup sends are best-effort. `qwp_ws_background_orphan_close_does_not_dial_next_slot` now asserts that the close finishes within 5s, which the `unwrap()` used to enforce implicitly. The blocked-send mock now fails if it expires instead of being released explicitly.
- `system_test/enterprise_e2e/c_client_sidecar.py`: the two `_readline` calls no longer pass `self.process.stdout`, matching the Enterprise sidecar's `_readline(timeout)` API.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket ingestion bookmark and marker restores across multiple interleaved tables.
  * Ensured rows and tables added after a rewind point are discarded while earlier data remains intact.
  * Improved rollback handling for open rows and different column types, including symbols, decimals, and geohashes.
  * Ensured buffers can be reused reliably after a rollback.

* **Documentation**
  * Clarified that clearing a buffer discards any stored marker or bookmark while preserving its capacity.

* **Tests**
  * Added regression coverage for multi-table rewinds, failed publications, and buffer reuse.


eugenels:
## The length-based rewind only unwinds one row per table

The diagnosis is right — `bookmark()` really was cloning everything already buffered — but the replacement mechanism is correct only for the workload that motivated it. That's why the benchmark and the full suite are green.

The doc comment (and PR body) states:

> `QwpWsTableBuffer::restore` truncates every column's data back to `row_count`

It doesn't. `restore` delegates to `QwpWsColumnBuffer::rollback_row`, which is a single-cell pop:

```rust
fn rollback_row(&mut self, row_idx: u32) {
    if self.last_written_row != Some(row_idx) {
        return;                                  // <- bails out entirely
    }
    if self.values.rollback_row(row_idx) { ... } // pops ONE cell
```

That was written for `rollback_current_row`, where exactly one uncommitted row exists so `last_written_row == row_count` holds by construction. A bookmark or marker can span any number of rows. Once two or more rows have been appended to a table after the mark, `last_written_row` is the *last* of them and never matches, the guard returns early, and **nothing is rolled back** — cells, slice data, symbol dictionary entries, `non_null_count` and decimal scale all survive. `restore` assigns `row_count = mark.row_count` regardless, so the table claims N rows while its columns hold N+k. The frame that goes out has a header declaring one row next to a symbol column emitting three cells, three timestamps, and the discarded symbol values promoted into the connection-global dictionary.

### Why the tests don't catch it

`test_bookmark_rewinds_every_table_it_spans` appends exactly one row per table after the bookmark — which is precisely the case where the single-cell pop is correct. It also asserts only on `row_count()`, a table-level counter the restore assigns directly, so it reads correct even when every column still holds the discarded data. Only the encoded bytes can detect this.

The downstream ingest path in the PR body is in the same working case (one row between mark and rewind), so those numbers are genuine.

### It's not only `bookmark()`

`capture_snapshot` / `restore_snapshot` also back `set_marker` / `rewind_to_marker`, exported as `line_sender_buffer_set_marker` / `line_sender_buffer_rewind_to_marker` and documented in `line_sender.h` as the way to "rewind rows" — plural, multi-row by design. Every C, C++ and Python caller that markers a batch and rolls it back on error is in the broken case.

### Regression test

Fails on `49f74eb`, passes on `main`. Three rows per table after the mark instead of one, both rewind surfaces, and an assertion on encoded bytes rather than `row_count()`. Goes next to the existing rewind-point tests in `questdb-rs/src/ingress/buffer/qwp.rs`:

```rust
    /// A rewind must unwind every row appended after the mark, not only the
    /// last one.
    ///
    /// `row_count()` alone cannot see a partial rewind: it is a table-level
    /// counter that the restore assigns directly, so it reads correct even
    /// when the column buffers still hold the discarded cells. The decisive
    /// check is the encoded frame -- a rewound buffer must serialise
    /// byte-for-byte like one that only ever held the surviving rows.
    #[cfg(feature = "_sender-qwp-ws")]
    #[test]
    fn qwp_ws_columnar_rewind_unwinds_every_row_after_the_mark() {
        // Written before the mark: must survive untouched.
        fn write_kept_rows(buf: &mut QwpWsColumnarBuffer) {
            buf.table("trades")
                .unwrap()
                .symbol("sym", "ETH-USD")
                .unwrap()
                .at(TimestampNanos::new(1_700_000_000_000_000_000))
                .unwrap();
            buf.table("quotes")
                .unwrap()
                .column_i64("bid", 10)
                .unwrap()
                .at(TimestampNanos::new(1_700_000_000_000_000_001))
                .unwrap();
        }

        // Written after it: several rows per table, so the rewind spans more
        // than one row of each; a column that did not exist at the mark; a
        // table that did not exist at all; and new symbol values, which must
        // not survive in the column dictionary either.
        fn write_discarded_rows(buf: &mut QwpWsColumnarBuffer) {
            for i in 0..3i64 {
                buf.table("trades")
                    .unwrap()
                    .symbol("sym", format!("BTC-USD-{i}"))
                    .unwrap()
                    .column_str("note", "added after the mark")
                    .unwrap()
                    .at(TimestampNanos::new(1_700_000_001_000_000_000 + i))
                    .unwrap();
                buf.table("quotes")
                    .unwrap()
                    .column_i64("bid", 200 + i)
                    .unwrap()
                    .at(TimestampNanos::new(1_700_000_002_000_000_000 + i))
                    .unwrap();
                buf.table("late_table")
                    .unwrap()
                    .column_i64("v", i)
                    .unwrap()
                    .at(TimestampNanos::new(1_700_000_003_000_000_000 + i))
                    .unwrap();
            }
        }

        // A fresh connection dictionary each time, so two buffers holding the
        // same rows encode identically, symbol delta section included.
        fn encoded(buf: &mut QwpWsColumnarBuffer) -> Vec<u8> {
            let mut scratch = QwpWsEncodeScratch::new();
            let mut global_dict = SymbolGlobalDict::new();
            buf.encode_ws_replay_message(&mut scratch, &mut global_dict, QWP_VERSION_1)
                .unwrap();
            scratch.message.clone()
        }

        let mut reference = QwpWsColumnarBuffer::new(127);
        write_kept_rows(&mut reference);
        let expected = encoded(&mut reference);

        // Both rewind surfaces share one capture/restore path.
        for api in ["marker", "bookmark"] {
            let mut buf = QwpWsColumnarBuffer::new(127);
            write_kept_rows(&mut buf);

            let bookmark = if api == "marker" {
                buf.set_marker().unwrap();
                None
            } else {
                Some(buf.bookmark().unwrap())
            };

            write_discarded_rows(&mut buf);
            assert_eq!(buf.row_count(), 11, "{api}: rows before the rewind");

            match bookmark {
                Some(bookmark) => buf.rewind_to_bookmark(bookmark).unwrap(),
                None => buf.rewind_to_marker().unwrap(),
            }

            assert_eq!(buf.row_count(), 2, "{api}: only the pre-mark rows survive");
            assert_eq!(
                encoded(&mut buf),
                expected,
                "{api}: a rewound buffer must encode exactly the pre-mark rows; \
                 discarded cells, symbols and columns must not reach the wire"
            );
        }
    }
```

### Suggested fix

Keep the length-recording idea, but push it down to **per column** rather than per table — the mark can't delegate to `rollback_row`, which only knows how to undo one row. Record at capture time, for each column:

- `cells.len()` and the backing `data.len()` for the slice-valued kinds
- `non_null_count` and `last_written_row`
- for symbol columns, `dict.len()` plus `symbol_cells_encoded_len` and `symbol_dict_encoded_len`
- for decimal columns, `decimal_scale`

`restore` then truncates each column to those lengths and rebuilds the symbol lookup with the existing `retain_local_ids_below`. It unwinds any number of rows by construction rather than by coincidence.

Capture becomes O(columns present) instead of O(1) per table, but stays independent of how many rows are buffered — which is the property the benchmark measures. The flat per-row cost across 100 → 50,000-row frames survives; only the per-mark constant grows, with schema width rather than buffer fill.

eugenels:
## Review summary (level 3, range `8ef0219f..092ce670`)

Full pass: change-surface map, callsite inventory across Rust, FFI, C header, C++ wrapper and Python shim, ten parallel review agents, and per-finding verification against source and base. The multi-row rewind bug from my earlier comment is fixed in `fc740e64` and pinned by the tests in `4c5cdcf1`; the three commits after it come out of this review.

### Fixed on the branch

- **Critical — test build broken without the WebSocket feature.** The two sender-level rewind tests called `Buffer::qwp_ws_with_max_name_len`, which only exists under `_sender-qwp-ws`, from an unconditionally compiled module. Plain `cargo test` passed because the default features bundle the WebSocket sender, but the TCP-only, HTTP-only and reader-only legs of `ci/run_all_tests.py` failed to compile the test target, which is why every cargo-test job on this PR was red. Gated in `1c8fc16b`; all three legs compile again.
- **Coverage — the byte oracle was blind to leftover cells in fixed-width kinds.** Bool, I8/16/32/64, F32/64 and Char encode by walking row indexes with a cursor, so a cell past the row count never reaches the wire and does not move `len()`. The sweep test now appends past the rewind and compares again (`a1450212`). Mutation-checked: disabling the I64 pop leaves every earlier assertion green and fails only the new one.
- **Coverage — rewind with an open row was untested for this buffer.** Rewind is not gated on the row boundary; the ILP and UDP buffers both had such a test, the buffer whose rewind was rewritten did not. Added in `092ce670`, for both surfaces and for the open row in a pre-mark table and in a post-mark one.

### Body updated

The "worth a second opinion" items are answered in the body: a successful flush calls `clear()`, which drops the snapshot, so nothing can rewind across it; the dirty marking is narrowed to the tables the rewind unwound and pinned by a test. The measured numbers predate the fix, which leaves the capture path they are dominated by untouched.

### Verified and cleared

- All 20 `rollback_current_row` error paths, both `restore` callers, the `Buffer` dispatch, the six FFI exports, the C header, the C++ wrapper copy/move paths and the Python shim were walked individually. No sender-internal code sets a marker or bookmark, so no flush or store-and-forward path was affected by the original bug.
- No lock is held across a rewind, the buffer never crosses to a drainer thread, `encode_ws_replay_message` takes `&self`, and the snapshot's auto-traits are unchanged.
- The counter arithmetic in `rollback_rows_from` can only go wrong at 2^32 non-null cells in one column, beyond any reachable buffer.

### Remaining, non-blocking

- **Geohash precision after a rewind that empties the column** stays at the value the discarded row pinned, where the clone restored the earlier one. Needs a precision change after the mark to reach; the row-level rollback path already behaved this way. Noted in the body as a known difference; a per-column mark for that one case is possible if the server turns out to reject an all-null geohash column at a mismatched precision.
- Decimal scale after a rollback no longer "heals" a mixed-scale column to the max surviving scale; it keeps the first-value pin, matching what the same buffer does with no rollback. Arguably a fix.
- Perf refinements, not regressions: `capture_snapshot` allocates the mark vector per call and could recycle it as the size-hint scratch does; `restore` could return early for a table it did not change.
- Style: the `QwpWsSnapshot` doc narrates the old implementation; the first commit's scope is `qwp/ws` where history uses `qwp`.

Test gate: full suite green with `sync-sender-qwp-ws,sync-sender-qwp-udp,sync-reader-qwp-ws`; TCP-only, HTTP-only and reader-only test targets compile; clippy clean on the touched crate.

bluestreak01 (changes requested):
Changes requested for the GEOHASH state-restoration regression at f530e04. The remaining findings are nonblocking. Full verified review: https://github.com/questdb/c-questdb-client/pull/191#issuecomment-5571547540

bluestreak01:
@nwoolmer — level-3 review of commit `f530e04c09af62881261d5487df59d3cf4e447c5`. **Changes requested** for the GEOHASH restoration regression below.

## Critical

### 1. Rewind preserves discarded GEOHASH precision — **in-diff**

**Location:** `questdb-rs/src/ingress/buffer/qwp.rs:2910–2917,4723–4728`

Reproduction:
1. Write `g` with precision 25, then clear the buffer.
2. Complete a row containing only `n`; capture a bookmark or marker.
3. Append a row containing `g` with precision 5.
4. Rewind and flush.

The surviving row now advertises precision **5**, not **25**. Appending repins the empty retained column (`4493–4495`), snapshots omit its previous precision, and encoding emits precision even for all-null columns (`5131–5142`).

Both rewind probes **fail at HEAD and pass against base**. This violates the restoration contract in `questdb-rs/src/ingress/buffer.rs:638–660`.

Untouched standalone and pooled publishers propagate the incorrect metadata:
- `questdb-rs/src/ingress/sender/qwp_ws_publisher.rs:141–156`
- `questdb-rs/src/ingress/column_sender/sender.rs:1565–1618`

Against an existing `GEOHASH(25b)` column, server source rejects this batch at `QwpWalAppender.java:569–575`. This consequence was source-verified at server revision `af65039997b883e767cd174da0a04869129daaa5`, **not live-server tested**.

**Fix:** Preserve and restore the prior precision without cloning payloads. Add clear/flush-reuse regression tests for both rewind APIs.

## Moderate

### 2. Shutdown test accepts an expired stall — **in-diff**

**Location:** `questdb-rs/src/tests/qwp_ws.rs:4403`; supporting lines `1668,4389–4397`.

The no-next-slot test has no elapsed-time assertion. Ignoring the release-send error lets it pass after the mock independently closes its stalled socket.

A temporary mutation that stopped subsequent slots but waited for socket expiry **passed HEAD in 10.02 seconds; the base test rejected it**.

**Fix:** Assert bounded close time and fixture liveness separately from best-effort cleanup. This is weakened test coverage, not a demonstrated production shutdown defect; neighboring tests retain timing checks.

### 3. Missing WS rewind integration coverage — **in-diff**

**Locations:** `questdb-rs/src/tests/sender.rs:613–717`; `questdb-rs/src/ingress/buffer/qwp.rs:10328–10334`.

The new public tests never publish, while the byte oracle resets the connection dictionary. Existing foreign-language marker tests use ILP/UDP; pooled marked-buffer tests exercise empty buffers.

**Fix:** Add a supported C/C++ WS regression and a nonempty keep/failed-publication → rewind → republication test using the same connection dictionary. This is a coverage request, not another proven product bug.

## Minor

- **Incomplete matrix — in-diff:** `questdb-rs/src/ingress/buffer/qwp.rs:10437–10487`. The sweep omits I8, I16, F32, LongArray, Decimal64 and Decimal128, plus cleared-schema starting states. Extend it; severity is reduced because these variants share tested rollback helpers.
- **Commit subjects — out-of-diff metadata:** All seven exceed the under-50-character convention in `.pi/skills/review-pr/SKILL.md:418–420`. Shorten the final squash subject, or individual subjects if retained.

None wholly dismissed. Coverage findings were narrowed; no silent-loss, FFI-abort, or current shutdown-bug claim is established.

## Summary

**Request changes: fix the GEOHASH restoration regression before merge.**

- **5 findings verified, 0 removed:** 4 in-diff, 1 metadata-only. Expanded caller review found propagation of finding 1, not another independent out-of-diff defect.
- **1,779 tests passed, 23 ignored.** Formatting, plain clippy, TCP/HTTP/reader-only test builds, and Rust example compilation passed.
- No local C++/Python or live-server execution.
- Reviewed HEAD: `f530e04c09af62881261d5487df59d3cf4e447c5`; comparison base: `8ef0219fbcec6f88ceb12e2f3a346ca6e3a8c91c`. Repository files were not changed by the review.

eugenels:
Follow-up to @bluestreak01's level-3 review of `f530e04c`. All four in-diff findings are addressed in seven commits, `10fe41eb`..`0d9f02c5`; the PR body has a new "Geohash precision" section replacing the old "Known difference" note.

## What the review found → what changed

| # | Finding | Commit | Change |
|---|---|---|---|
| 1 | **Critical** — rewind preserves a discarded GEOHASH precision | `10fe41eb` | Confirmed as a regression against the clone-based rewind. `QwpWsColumnValues::Geohash` now carries `displaced_precision_bits`: `append_geohash` saves the old precision when it repins an empty column, and the rollback puts it back when it empties the column again. One value is exact because a repin only happens while empty and a rollback can only empty a column that was empty at the mark. Verified red at `f530e04c` (precision 5 where 25 is expected), green after. No change to the capture path; enum size unchanged (104 bytes). |
| 2 | **Moderate** — shutdown test accepts an expired stall | `b8c874ca`, `0d9f02c5` | `qwp_ws_background_orphan_close_does_not_dial_next_slot` now bounds `close_drain` (< 5s against the mock's 10s park) and, since the pool detaches a worker it cannot stop at ~3s, also reopens the stalled slot after the close — a detached worker keeps the slot lock. |
| 3 | **Moderate** — missing WS rewind integration coverage | `e6e10eab`, `c00011fd` | Rust: `qwp_ws_rewind_after_a_failed_publication_republishes_only_surviving_rows` — `flush_and_keep`, bookmark, oversized flush fails (rolls the connection dictionary back), rewind, republish; asserts `delta_start = 1`, `delta_count = 1`, only `keep-d` ships, no `drop-` symbol, row count 2. C++: `borrowed_sender publishes only the rows a bookmark rewind kept` on the pooled buffer, which is the supported C/C++ route to a QWP/WS row buffer (`line_sender::new_buffer` rejects WS). Both fail under a single-row-unwind mutation. |
| 4 | **Minor** — incomplete kind matrix | `9c5e6a8b` | Sweep gains I8, I16, F32, LongArray, Decimal64, Decimal128, and runs every case a second time from a cleared buffer. The cleared start independently catches finding 1. |
| 5 | **Minor** — commit subjects > 50 chars | — | Not rewriting pushed history. The seven new subjects are all ≤ 50. |

## Second pass over the fix

I ran a level-3 review over the seven commits themselves (correctness, panic surface under the FFI crate's `panic = "abort"`, FFI, concurrency, resources, performance, coverage, API, cross-context callers, fresh-context adversarial; per-finding verification with mutation runs). Result: no Critical, one Moderate coverage gap, seven Minors — all addressed in `f314d613` and `0d9f02c5`:

- **Moderate gap:** the precision restore is reached from two places — bookmark/marker rewind and the mid-row error rollback — but only the first had a test; the existing failed-row geohash test stayed green with the restore removed. Added `qwp_ws_columnar_failed_row_restores_the_precision_it_repinned`, red under both fix-removal mutations.
- The failed-publication test's doomed set was 256 rows, so a rewind that unwound only the last row still tripped the size cap on republish and the dictionary assertions never ran. Now two rows that each fit under the cap alone; under the mutation it fails on `delta_count` (2 vs 1).
- The doomed flush is asserted by message (`exceeds maximum configured allowed size`), not just `InvalidApiCall`, so the size cap is provably the guard that fired.
- Three comments that overstated what the code does were corrected.

Dismissed after verification: a stale `displaced_precision_bits` after `clear()` (always overwritten by the next pin before any pop can read it); precision 0 being restorable (a column present-but-empty at a mark must have been cleared, and one whose first append failed is truncated by the rollback); flakiness of the 5s bound (healthy close is ~2.5s, the same margin the three sibling tests already run with).

## Verification

`cargo test --features almost-all-features`: 2146 + 79 + doctests pass. `test_column_sender` 25/25, `test_line_sender` 94/94. `cargo fmt`, plain clippy (no warnings in touched hunks), `clang-format` clean. No live-server run; the server-side rejection in finding 1 is source-verified only, as in the original review.

One gap noted but out of scope here: the Python system tests bind only `set_marker`/`rewind_to_marker`, not the bookmark API, and have no WS rewind test.

eugenels:
Level-3 review of `0d9f02c5` (ten review agents, per-finding verification, eleven mutants run against the rewind tests in a throwaway worktree, full `questdb-rs` lib suite and the C++ `test_column_sender` / `test_line_sender` binaries). Result: no Critical, no code defect. One Moderate coverage gap and eight Minors, all addressed in nine commits, `d102ab70`..`e112b67e`.

## Findings → commits

| # | Finding | Commit | Change |
|---|---|---|---|
| 1 | **Moderate gap** — the `size_hint.mark_dirty` a rewind performs was pinned by nothing: every rewind test appended again afterwards, which re-dirtied the tables. Removing the mark survived the suite while `len()` reported the pre-rewind size until the next append. | `37384295` | The column-kind sweep and the public `test_rewind_discards_every_row_after_the_rewind_point` now fold the hint before rewinding. Both fail under the mutant (stale 253 vs 136, 930 vs 626). |
| 2 | `restore` computed `changed` and then walked every column of every surviving table anyway: O(all columns) per rewind. | `d102ab70` | Early return when the row and column counts match and no row is open; `debug_assert!(!mark.in_progress)` states the boundary precondition the predicate relies on. The early-return variant passed the whole suite as a mutant before being adopted. |
| 3 | `capture_snapshot` allocated a marks vector per bookmark, the last allocation on a bookmark-per-row path. | `676ed43c` | The previous snapshot's vector is taken back and refilled. |
| 4 | `retain_local_ids_below` swept the whole symbol lookup, O(distinct values), for one or a few popped ids. | `6a4acfb0` | `QwpWsLocalSymbolLookup::remove(hash, local_id)` drops the one slot; the popped entry's bytes are rehashed before the arena truncate. Collision test extended to cover both bucket shapes down to an empty map. |
| 5 | `restore_snapshot` comment claimed "no per-column state is needed", which the geohash fix disproved; `QwpWsSnapshot` doc restated the commit history. | `736c22e2` | Trimmed. |
| 6 | No websocket test covered rewinding after `clear()`. | `c09ab2b3` | `qwp_ws_columnar_clear_drops_the_rewind_point`: the marker reports "No marker set", a bookmark handle "Bookmark is stale". |
| 7 | The decimal "keep pinned" block wrote "5.6" after a scale-1 survivor, so an erroneous unpin re-pinned scale 1 and encoded the same bytes (the sweep still caught an always-unpin mutant, so suite-level coverage was intact). | `921e181e` | Pins scale 2 with "1.25"; an unpinned column would let "5.6" pin scale 1, which "1.25" cannot encode at. |
| 8 | Four verbatim copies of the best-effort-release note; `first_table_row_count` re-walked `first_table_column_count`'s prefix. | `6f274532` | One note on the parking mock; shared `first_table_body_pos`. |
| 9 | C/C++/FFI docs for `clear()` did not say the rewind point is dropped, while the Rust doc does (pre-existing). | `e112b67e` | One sentence on each. |

PR body: the flush paragraph now names the actual errors (`rewind_to_marker` → no marker set, bookmark handle → stale) and notes that `flush_and_keep`, successful or not, leaves the bookmark valid.

## Dismissed after verification

- `last_written_row` clearing "only incidentally tested": seven tests kill that mutant.
- Non-null counter drift after a failed `increment_non_null`: needs 2^32 non-null cells in one column; base was worse.
- Retained capacity after a rewind: matches the documented `clear()` contract; base freed memory only as a side effect of cloning.
- A stale symbol-lookup entry is unobservable: `get` validates every id against the dictionary. Finding 4 is a cost fix, not a correctness one.

## Gates on `e112b67e`

- `cargo test --lib` (questdb-rs): 1700 passed.
- `test_column_sender`: 25 cases / 89 assertions; `test_line_sender` marker and bookmark cases: 6 / 117.
- `cargo fmt` clean; `cargo clippy --tests` shows only pre-existing warnings on untouched lines.

Not changed: commit scopes and subject lengths on the already-pushed history, and the Python system tests, which have no websocket bookmark bindings at all.

bluestreak01 (changes requested):
@nwoolmer

Reviewing PR #191 at level 3, head c7470b9763e8a1d3adaa04af7cc5e3ebd2be9683, base 621a6baffdc1611474810e4c9e04cc575722f874.

## Critical

### 1. Large-symbol row rejection

**Problem:** Rejecting rows with large symbols takes almost twice as long.  
**Net impact:** QWP/WS large-symbol error handling adds approximately 1 ms per rejected MiB.  
**Evidence:** `falsify-symbol-summary.csv`, reviewed HEAD versus BASE: **1.996 ms versus 1.009 ms**.

**In-diff:** `questdb-rs/src/ingress/buffer/qwp.rs:4754–4755`.

Commit a symbol and integer column; append a new 1-MiB symbol to the existing column, then write a string into the integer column. The resulting `InvalidApiCall` triggers rollback, which hashes the symbol bytes again—one additional byte-linear pass per rejected symbol, without requiring a bookmark.

**Net-negative for this producer:** the slowdown persists with warmed capacity and 4,096 retained symbols. Avoided dictionary scans benefit short symbols but do not offset this case. Row preservation and subsequent recovery passed on both revisions.

**Fix:** reuse the insertion hash for removal. A compiling scratch prototype restored the operation to approximately **1.010 ms**; its cached-hash approach costs eight additional bytes per dictionary capacity slot.

### 2. Sustained large-suffix rewinds

**Problem:** Repeated large-batch rewinds regress local buffer-build latency.  
**Net impact:** QWP/WS i64 batch builders lose approximately 4% over sustained append/rewind cycles.  
**Evidence:** `falsify-bulk12-total-summary.csv`, reviewed HEAD versus BASE: **303.54 ms versus 292.21 ms**.

**In-diff:** `questdb-rs/src/ingress/buffer/qwp.rs:4085–4089`.

Retain one committed i64 row, bookmark, append one million rows, rewind, and repeat. Restoration now visits every discarded cell. The twelve-cycle measurement includes capture, append and rewind, reusing the same buffer throughout.

**Net-negative for this sustained producer:** cheaper capture and retained capacity were included. Shorter three-cycle runs improved, but twelve-cycle runs remained 3.9% slower. This is additional per-discarded-cell work, not quadratic overall complexity.

**Fix:** locate the suffix boundary and truncate fixed-width cells in bulk, preserving the single-row fast path. The scratch i64 prototype reduced the same total to **282.30 ms** and passed the full Rust library suite.

## Moderate

### 3. Singleton cleanup coverage

**Problem:** Singleton lookup cleanup lacks an effective regression assertion.  
**Net impact:** QWP/WS rejected-row lookup growth can escape existing wire assertions.  
**Evidence:** HEAD with singleton removal disabled passes **2,150 tests**; a diagnostic retains **1,001 buckets**, versus **one** on unmodified HEAD and BASE.

**In-diff:** `questdb-rs/src/ingress/buffer/qwp.rs:2379–2391,8080–8114`.

The collision test exercises `Many`, not singleton removal. Repeated fresh-symbol writes followed by a wrong-type column error expose this blind spot. Dictionary bounds/byte checks preserve correct wire output; `clear`/Drop reclaim the map. **Current HEAD cleanup is correct**—this is a coverage gap, not an observed production leak.

Add a singleton-removal/bucket assertion to the existing unit test. That extension passed on HEAD and failed under the mutation, without timing or networking.

## Coverage and validation

- **Test gate: passes; one admitted Moderate coverage gap.** Recorded search: `QwpWsLocalSymbolLookup|lookup\.buckets` across Rust/C++/Python sources. Missing assertion: live bucket reclamation. Tier: Rust library PR CI.
- **2,150 Rust tests passed**, 23 ignored; **119 C++ cases passed**. Rust formatting, plain clippy, doctests and TCP-only/HTTP-only/reader-only compile checks passed. Local C++ formatting was unavailable.
- CI, including **TestVsQuestDBMaster**, passed. That supplies general live integration coverage; the new WS rewind scenarios were exercised by unit/mock tests, not a dedicated live-server rewind test.
- FFI dev/release uses **panic=abort** (`questdb-rs-ffi/Cargo.toml:73–77`); standalone core/tests unwind. Overflow checks default **on** in dev/test, **off** in release.
- **Submodules:** no pointer changes.

## Summary

**Verdict: request changes.** Correctness gate fails on the performance findings; test gate passes.

**2 Critical, 1 Moderate, 0 Minor—all in-diff.** No out-of-diff-breakage findings.

Paired release probes used Rust 1.91.1, QWP/WS with `ring-crypto` and `tls-webpki-certs`, and pinned CPU execution. Sources, commands, raw results and remedy prototype: `/tmp/pr191-review.KWRGVm/`.

Primary and scratch worktrees are clean. Nothing was committed or published to GitHub.

mtopolnik:
Reviewing PR #191 at level 3, head 3d6c0c7610c07871f1993751b7d1ff46e99c5113, base 4123776a64642d932924ea2a164650a18dc394dd.

Review performed by **sol/xhigh**.

**Verdict: approve with comments.** I found no demonstrated bug in the buffer change. I have two test comments:

1. **The orphan shutdown test may check the wrong slot.** It creates `orphan-a` and `orphan-b`, then tries to reopen only `orphan-a`. The code does not guarantee which orphan is processed first. If `orphan-b` is the stalled one, the test can pass even while its lock remains held. The [test](https://github.com/questdb/c-questdb-client/blob/3d6c0c7610c07871f1993751b7d1ff46e99c5113/questdb-rs/src/tests/qwp_ws.rs#L4897) should check both slots or record which one stalled.

2. **A rewind from an empty, reused WebSocket buffer is untested.** After `clear()`, the buffer keeps its table structure. A caller can set a bookmark at zero rows, add rows, rewind, and add replacement rows. The [current test](https://github.com/questdb/c-questdb-client/blob/3d6c0c7610c07871f1993751b7d1ff46e99c5113/questdb-rs/src/ingress/buffer/qwp.rs#L10845) starts with at least one row. Adding a zero-row case would check that discarded values cannot appear in the replacement frame. I found no failure in the current code.

Local Rust rewind tests and the rebuilt C++ rewind test passed. `TestVsQuestDBMaster` passed. The separate Enterprise e2e check failed, but its logs were inaccessible, so I could not determine whether that failure relates to this PR.

Large one-time rewinds are slower than before, but repeated use of the same buffer is faster because it keeps memory ready for reuse. No repository files were changed.

bluestreak01 (approved):
Reviewing PR #191 at level 3, head 33901e2aea798852148667a00c5dfbae46614a07, base 4123776a64642d932924ea2a164650a18dc394dd.

**Verdict: approve.** I found no blocking problems. One cosmetic item is optional.

### PR title and description
- The title is valid Conventional Commits, and the description explains why the change matters to users. There is no public API change: the C, C++ and FFI edits only update the `clear()` comments.
- **Description drift:** two sentences describe an earlier version of the code:
  - "the column loops until nothing pops": fixed-width columns are now cut off in one step (`650f84b0`).
  - "There is no early-return guard left": `restore` now returns early for tables the rewind didn't change (`d102ab70`).

  Update those sentences before merge.

### Critical
None.

### Moderate
None.

### Minor
- **Problem:** `pop_value_cell_from` returns an `Option<bool>` that its only caller ignores.
- **Net impact:** None for users; the leftover return value makes the code harder to read.
- **Evidence:** `questdb-rs/src/ingress/buffer/qwp.rs:2712-2714`. Its only caller is at `:2729`. Since `2c4c2a7a`, fixed-width cells no longer go through `pop_cell_from`.
- **Fix:** In the single-cell branch of `discard_fixed_width_suffix`, call `cells.pop()` directly and delete the helper. The branch has already checked that the last cell is at or past the rewind row.

### Adjacent findings
None.

### Coverage map
The test gate passes, with no coverage gaps worth raising.
- **What I ran on head `33901e2a`:**
  - The full `questdb-rs` library suite with `almost-all-features` passes: 2179 passed, 23 ignored.
  - The C++ `test_column_sender` passes 25/25, including the new bookmark-rewind case, which checks the published frame.
  - `cargo fmt --check` and plain `cargo clippy --tests` are clean.
  - The test builds still compile with only the TCP sender, only the QWP/UDP sender, or only the reader enabled.
- **Fuzz test:** I wrote a throwaway test and ran it in a scratch worktree.
  - It writes random rows across 18 column kinds, including symbol, decimal and geohash. It mixes in type-mismatch rows that fail, markers, bookmarks, rewinds (some mid-row) and `clear()` calls.
  - After every step it compares the buffer with one holding only the rows that should survive. It checks encoded bytes, `len()`, `recompute_len_slow()`, `row_count()` and `is_empty()`.
  - It passed 2000 seeds × 400 steps.
  - It fails, as it should, when I disable the geohash precision restore or make the decimal scale reset unconditional.

### Summary
- **Findings:** none blocking; one Minor, in the diff; nothing broken outside the diff.
- **Panic and overflow settings:** `questdb-rs-ffi` sets `panic = "abort"` in both dev and release (`questdb-rs-ffi/Cargo.toml:73-77`). Overflow checks are on in dev/test and off in release.
  - The new `debug_assert!(!mark.in_progress)` can't fire: rewind points, and the row mark `table()` takes, are only captured between rows.
  - The other new indexing, `expect` and subtraction sites rely on buffer invariants that callers can't break through the API.
- **Submodules:** the PR doesn't move any submodule pointers.
- **Protocol:** the wire format doesn't change, only how the buffer is rewound. The Vs QuestDB master job was still pending on this head when I checked.
- **Performance:** capturing a rewind point is now O(tables) instead of O(size of the buffer). A rewind costs O(tables + columns of the tables it changed + discarded cells). A mid-row error rollback no longer rebuilds lookups or re-sums symbol sizes every time. Rewinding many rows of string, binary or array columns still removes cells one at a time. That costs no more than appending those rows did, so it isn't a regression.

PR #192, opened 2026-08-26
chore(build): run the macOS CI legs on Apple Silicon agents

mtopolnik:
## What broke

The hourly fuzz pipeline has failed on its macOS leg every run since
2026-08-25 15:00 UTC — always on the same commit, always before a single test
started. QuestDB refused to boot:

```
io.questdb.std.ex.FatalError: QuestDB does not support x86-64 (Intel) macOS,
no native library is bundled for it. Supported platforms: macOS aarch64,
Linux x86-64, Linux aarch64, Windows x86-64.
	at io.questdb.std.Os.loadLib(Os.java:348)
```

which `system_test/fixture.py` reports as `RuntimeError: QuestDB died during
startup.` ([build 264770](https://dev.azure.com/questdb/questdb/_build/results?buildId=264770&view=results))

Nothing changed on our side. Every one of these jobs clones questdb master and
builds the server from source, and questdb/questdb#7555 landed at 14:44 UTC
that same day: it deletes the `darwin-x86-64` dylibs from the jar and moves
questdb's own macOS test legs to Apple Silicon. Azure's `macos-latest` image is
Intel — the failing run resolved its JDK under `/usr/local/opt`, the x86-64
Homebrew prefix — so the jar we build no longer carries a library that agent
can load. The last green run was 14:00 UTC, the first red one 15:00 UTC.

The PR-time pipeline had not run since 13:00 UTC that day, so it has not gone
red yet, but it would fail identically on the next PR.

## What changed

Every macOS leg in both pipelines now asks for `macOS-15-arm64`, the image
questdb/questdb switched to. That covers the three jobs that start a server
(`NativeTests`, and `TestQwpWsFuzz` in each pipeline) and also the two that
only run cargo (`RustTests`, `TestQwpEgressFuzz`). aarch64 is the only macOS
build the client ships, so there is no reason to keep testing the other one.

Finding the JDK moves into a new template, `ci/templates/resolve_jdk25.yaml`,
shared by the three server-building jobs instead of being copied into two of
them. On macOS it reads `$JAVA_HOME_25_ARM64`, which the arm64 image sets, and
falls back to `/usr/libexec/java_home -v 25` and then to a Homebrew install; on
Linux and Windows it reads `$JAVA_HOME_25_X64` as before. It also stops with a
clear message when the agent is not arm64 — the fixture starts the server from
`$JAVA_HOME/bin/java`, so an Intel JDK there is exactly what produced the
`FatalError` above several minutes into the run.

Two supporting changes fall out of the image swap:

- `NativeTests` read the JDK straight from `$(JAVA_HOME_25_X64)`, which does
  not exist on an arm64 image, so it now uses the shared template and
  `$(JAVA_HOME)`.
- The two steps in `compile.yaml` that install the Python and Arrow
  dependencies chose between Homebrew and pip by comparing the image name to
  `macos-latest`. Renaming the image would have quietly sent the macOS legs
  down the pip branch and skipped `apache-arrow`, so they now check the agent's
  operating system instead.

The second commit bumps the `questdb` submodule to `4777a4e7`, the commit that
dropped Intel macOS support, so a local run reproduces what CI sees.

## Validation

- Both pipelines expand and validate against the real Azure YAML parser via the
  `pipelines/{id}/preview` API (`previewRun: true`, so nothing was queued).
- `macOS-15-arm64` is confirmed live in this org: questdb's own `macwin` build
  264683 logs `image=macOS-15-arm64 arch=arm64 os=15.7.7` and resolves JDK 25 at
  `/Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/25.0.3-9.0/arm64/Contents/Home`
  — the first branch of the new template, so the Homebrew fallback never runs on
  the happy path.
- The resolution script passes `shellcheck` and was executed for all four paths:
  arm64 with the variable set, an Intel agent (fails with the intended message,
  never reaching brew), Linux/Windows, and a resolved directory with no
  `bin/java`.

Not proven without a real run: that the Maven build and the fuzz suite complete
end to end on that image. questdb's own macOS legs already build the same jar
there, so the residual risk is narrow — Homebrew `apache-arrow`/`numpy` on
arm64, and wall-clock.

## Summary by CodeRabbit

- **Chores**
  - Updated macOS CI and fuzz-testing jobs to use consistent ARM64 macOS 15 environments.
  - Improved Java 25 setup across supported build agents, with validation and clearer failure reporting.
  - Standardized platform detection for dependency installation.
  - Updated the bundled QuestDB component to a newer revision.

### DIFF
--- a/ci/run_fuzz_pipeline.yaml
+++ b/ci/run_fuzz_pipeline.yaml
@@ -13,6 +13,11 @@ pr: none
 # server). Equivalent PR-time jobs live in run_tests_pipeline.yaml
 # (TestQwpWsFuzz / TestQwpEgressLiveServerFuzz); this pipeline is the
 # always-on stability watchdog and the source of Slack #builds alerts.
+#
+# The mac legs pin macOS-15-arm64. The plain macOS-14/15/26 Azure images are
+# Intel x86-64, and QuestDB bundles a macOS native library for aarch64 only,
+# so the server refuses to start on an Intel agent. This mirrors the image
+# questdb/questdb's own macOS legs run on.
 schedules:
   - cron: "0 * * * *"
     displayName: Run QWP/WS + QWP egress fuzz every hour
@@ -30,7 +35,7 @@ stages:
         strategy:
           matrix:
             mac:
-              imageName: "macos-latest"
+              imageName: "macOS-15-arm64"
             windows-2022:
               imageName: "windows-2022"
         pool:
@@ -42,30 +47,7 @@ stages:
             lfs: false
             submodules: false
           - template: templates/compile.yaml
-          # Same JDK 25 dance as the PR-time fuzz job in
-          # run_tests_pipeline.yaml — Linux/Windows hosted agents ship JDK
-          # 25 at $JAVA_HOME_25_X64; macos-latest does not, so we brew it
-          # in. Resolution stays in a single bash step so the Windows
-          # leg's cmd.exe quote handling doesn't leave a stray `"` on
-          # the resolved path.
-          - bash: |
-              set -e
-              if [[ "$AGENT_OS" == "Darwin" ]]; then
-                brew install openjdk@25
-                JDK_HOME="$(brew --prefix openjdk@25)/libexec/openjdk.jdk/Contents/Home"
-                if [ ! -d "$JDK_HOME" ]; then
-                  JDK_HOME="$(/usr/libexec/java_home -v 25 2>/dev/null || true)"
-                fi
-                if [ ! -d "$JDK_HOME" ]; then
-                  echo "JDK 25 not found after brew install" >&2
-                  exit 1
-                fi
-              else
-                JDK_HOME="$JAVA_HOME_25_X64"
-              fi
-              echo "Resolved JDK 25: $JDK_HOME"
-              echo "##vso[task.setvariable variable=JAVA_HOME]$JDK_HOME"
-            displayName: "Resolve JDK 25"
+          - template: templates/resolve_jdk25.yaml
           - template: templates/clone_questdb.yaml
           - template: templates/maven_settings.yaml
           - task: Maven@3
@@ -231,7 +213,7 @@ stages:
             linux:
               imageName: "ubuntu-latest"
             mac:
-              imageName: "macos-latest"
+              imageName: "macOS-15-arm64"
             windows-2022:
               imageName: "windows-2022"
         pool:
--- a/ci/run_tests_pipeline.yaml
+++ b/ci/run_tests_pipeline.yaml
@@ -74,12 +74,18 @@ stages:
           #   displayName: "Publish build directory"
       # mac/windows: run cargo vs native (build + C++ + integration) as two
       # parallel jobs to cut the slow legs' wall-clock. linux stays in RunOn.
+      #
+      # Every mac leg in this file pins macOS-15-arm64. The plain
+      # macOS-14/15/26 Azure images are Intel x86-64, and QuestDB bundles a
+      # macOS native library for aarch64 only, so a job that starts the server
+      # cannot run on an Intel agent. Keeping the cargo-only legs on the same
+      # image tests the one macOS architecture the client ships against.
       - job: RustTests
         displayName: "cargo tests (mac/windows)"
         strategy:
           matrix:
             mac:
-              imageName: "macos-latest"
+              imageName: "macOS-15-arm64"
             windows-msvc-2022:
               imageName: "windows-2022"
             windows-msvc-2025:
@@ -99,7 +105,7 @@ stages:
         strategy:
           matrix:
             mac:
-              imageName: "macos-latest"
+              imageName: "macOS-15-arm64"
             windows-msvc-2022:
               imageName: "windows-2022"
             windows-msvc-2025:
@@ -118,6 +124,7 @@ stages:
           - script: python3 ci/run_all_tests.py cpp
             displayName: "C++ tests"
           # Build QuestDB master for run_all_tests.py's `--repo ./questdb`.
+          - template: templates/resolve_jdk25.yaml
           - template: templates/clone_questdb.yaml
           - task: Maven@3
             displayName: "Compile QuestDB"
@@ -125,11 +132,11 @@ stages:
             inputs:
               mavenPOMFile: "questdb/pom.xml"
               javaHomeOption: "Path"
-              jdkDirectory: "$(JAVA_HOME_25_X64)"
+              jdkDirectory: "$(JAVA_HOME)"
               options: "-DskipTests -Pbuild-web-console$(CLIENT_PROFILE)"
           - script: python3 ci/run_all_tests.py integration
             env:
-              JAVA_HOME: $(JAVA_HOME_25_X64)
+              JAVA_HOME: $(JAVA_HOME)
             displayName: "Integration tests"
           - task: ArchiveFiles@2
             displayName: "Compress QuestDB server log on failure"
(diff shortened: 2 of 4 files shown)


mtopolnik:
## Review (level 0)

CI-YAML + a submodule pointer; no Rust/FFI surface, so the panic/overflow contract is untouched.

---

## Moderate

### 1. The shared template's non-Darwin branch silently sets an empty `JAVA_HOME`

**In-diff** — `ci/templates/resolve_jdk25.yaml:39-43`

```bash
else
  JDK_HOME="$JAVA_HOME_25_X64"
fi
echo "Resolved JDK 25: $JDK_HOME"
echo "##vso[task.setvariable variable=JAVA_HOME]$JDK_HOME"
```

There is no `set -u` and no existence check on this path. Executed proof:

```
$ env -u JAVA_HOME_25_X64 AGENT_OS=Linux bash jdk.sh
Resolved JDK 25: 
##vso[task.setvariable variable=JAVA_HOME]
exit=0
```

The step reports success and publishes `JAVA_HOME=""`. Downstream, `Maven@3` gets `jdkDirectory: ""` with `javaHomeOption: Path`, and `system_test/fixture.py:281-289` treats an empty `JAVA_HOME` as falsy and falls through to `shutil.which('java')` — whatever JDK happens to be on PATH, not necessarily 25. That is the same failure shape this PR exists to kill: a misresolved JDK surfacing minutes later as a confusing runtime error instead of at the resolution step.

The Darwin branch already has exactly the guard that catches this (`if [ ! -x "$JDK_HOME/bin/java" ]`, lines 33-37). The `else` branch doesn't get it.

Attribution: for the two `TestQwpWsFuzz` jobs, base (`8ef0219f`, `ci/run_tests_pipeline.yaml:420` / `ci/run_fuzz_pipeline.yaml:65`) has the identical unguarded assignment — pre-existing there. For **`NativeTests` it is new**: base passed `jdkDirectory: "$(JAVA_HOME_25_X64)"` directly, so a missing variable left the macro unexpanded and Maven failed loudly on a literal `$(JAVA_HOME_25_X64)` path. Head converts that loud failure into an empty value.

Fix — hoist the check out of the `if`, so it covers all three OS families:

```bash
  else
    JDK_HOME="$JAVA_HOME_25_X64"
  fi
  if [ ! -x "$JDK_HOME/bin/java" ]; then
    echo "JDK 25 not found on this agent (JDK_HOME='$JDK_HOME')" >&2
    exit 1
  fi
```

---

## Minor

### 2. `JAVA_HOME: $(JAVA_HOME_17_X64)` no longer resolves on the mac legs

**In-diff (consequence)** — `ci/templates/compile.yaml:41-46`

Both cmake steps set `JAVA_HOME: $(JAVA_HOME_17_X64)`. The arm64 images expose `JAVA_HOME_17_ARM64`, so on `macOS-15-arm64` Azure leaves the macro unexpanded and cmake runs with `JAVA_HOME` literally equal to the string `$(JAVA_HOME_17_X64)`.

It's inert: the only `find_package(Java)` in the tree is `CMakeLists.txt:479-482`, gated behind `option(QUESTDB_SYSTEM_TESTING "Run system tests" OFF)` (line 473), and nothing in `ci/` turns that on. But it is now dead and actively misleading — a future reader will assume the mac build has a JDK 17 pinned. Drop the `env:` block, or gate it.

### 3. The comment at `compile.yaml:24-26` reads as a blanket rule the file doesn't follow

**In-diff** — `ci/templates/compile.yaml:24-26` vs `13`, `17`

The new comment says to key on agent OS "not the image label", but the two steps directly above it still use `eq(variables['imageName'], 'ubuntu-latest')`. That's correct — those steps must *not* run on the self-hosted `hetzner-incus` Linux agents, so `Agent.OS` is genuinely the wrong key there. Say so in one clause, otherwise the comment reads as an unapplied rule.

### 4. Second commit message is contentless

`05743a11 "Update questdb module"`, empty body. It should name the bump (`3c1c3382` → `4777a4e7`) and state the blast radius: **CI never reads this pointer** — `ci/templates/clone_questdb.yaml:21` does `git clone --depth 1 https://github.com/questdb/questdb.git`, and every job checks out with `submodules: false`. The bump only affects local reproduction, which is what the PR body says but the commit doesn't.

### 5. `doc/COMPATIBILITY.md:40-41` is now an incomplete coverage statement

**Out-of-diff.** "CI covers GCC or Clang on Linux, Apple Clang on macOS, and the MSVC toolsets…" — after this PR that is Apple Clang on **arm64** macOS only. Not false as written (no architecture is claimed), but if the line is meant as the coverage contract, qualify it.

---

## Coverage gaps

None Critical or Moderate. The template's three macOS fallback branches (`java_home -v 25`, the Homebrew install, the `bin/java` failure) are only reachable on a real agent; the PR body records manual execution of all four paths, which is the practical ceiling here. Accepted.

- **`$(uname -m)` collides with Azure macro syntax** — Azure leaves unresolvable `$(...)` macros literal, and base already depended on this (`$(brew --prefix openjdk@25)` at the old `run_tests_pipeline.yaml:414`), green until 2026-08-25 14:00.
- **`variables['Agent.OS']` unavailable in a step condition** — set before steps run; the repo already uses it at `ci/run_tests_pipeline.yaml:47`.
- **The `compile.yaml` condition flip changes the `hetzner-incus` legs** — old `ne(imageName,'macos-latest')` and new `ne(Agent.OS,'Darwin')` both select the pip branch on Linux. No delta.
- **Dropping all x86-64 macOS legs leaves a shipped artifact untested** — this repo publishes no macOS binaries (no release pipeline under `ci/`). The only arch-gated code is `questdb-rs/src/ws/mask.rs`: its `target_arch = "x86_64"` path (lines 196-263) stays covered by the Linux/Windows x64 legs, and its `target_arch = "aarch64"` path (line 266) gains its first CI coverage here. macOS-gated code (`ws/nosigpipe.rs:67`, `ingress/sender/qwp_ws_sfa_segment.rs:1095`) stays covered.
- **`NativeTests` runs `run_all_tests.py cpp` before the new JDK step** — `ci/run_all_tests.py:60` `run_cpp_tests()` starts no server; only `run_integration_tests()` (line 90) reads `JAVA_HOME`, and it runs after the template.
- **The submodule bump changes what CI builds** — it doesn't; see finding 4.

---

## Summary

**Approve with comments.** The diagnosis is verified: `4777a4e7` is `build: drop x86-64 (Intel) macOS support and run macOS tests on Apple Silicon (#7555)`, the pipelines clone questdb master, and `macos-latest` on Azure is Intel. The fix targets the right thing, and the `uname -m` assertion converts a several-minute-deep `FatalError` into an immediate one. Nothing blocking.

- **Test gate:** no production code changed. `python3 ci/check_docs.py` passes (12 files, 85 local links); all four touched YAML files parse; the extracted `resolve_jdk25` script passes `shellcheck -s bash`. The claim that only a real Azure run proves the Maven build and fuzz suite complete on that image stands — it could not be verified here either.
- **Submodule provenance:** `questdb` 3c1c3382 → 4777a4e7 = **UPSTREAM-SYNC** (`git branch -r --contains 4777a4e7` lists `origin/HEAD -> origin/master`). Contents out of scope; integration reviewed and it is a no-op for CI.
- **Tradeoff:** x86-64 macOS loses all CI coverage. Justified — no arch-unique code lives in that intersection, and the aarch64 SIMD masking path in `ws/mask.rs` gains coverage it never had.
- **5 findings admitted, 6 candidates omitted as false positives.** 3 in-diff, 1 out-of-diff, 1 commit-message.
- **Admitted coverage gaps: 0.**

bluestreak01 (approved):

PR #193, opened 2026-08-26
fix(ingress): hold a GEOHASH value to its claimed precision at encode time

mtopolnik:
Split out of #189, which bundled this with an unrelated dispatcher lost-wakeup
fix. No code changes: commit `79e3ebb9` cherry-picked onto `main`. #189 now
carries only the dispatcher fix, which its title and body already described.

## The problem

The Arrow and NumPy column encoders truncate a GEOHASH value to the low
`ceil(bits/8)` bytes. A geohash's high bits are its coarse position
(`GeoHashes.java:215` interleaves and shifts right by `64 - bits`; `widen` at
`:316` lowers precision with a right shift), so a value wider than the declared
precision does not arrive as an error — it arrives as a well-formed geohash for
somewhere else entirely.

## The change

Every non-null value is held to the claimed precision at the point of
truncation, naming the column and the row.

- **Arrow** (`arrow_batch.rs`): the four hand-rolled per-dtype loops collapse
  into one `write_rows!` macro that range-checks before truncating.
  Error: `ArrowIngest`.
- **NumPy** (`numpy_wire.rs`): `check_geohash_range::<SRC>` runs as a pass ahead
  of the emit path so the whole-column memcpy stays available. `emit_into_wire`
  gains a `column: &str` so the message can name the column.
  Error: `InvalidApiCall`.

## Behaviour change

User-visible and reachable from Rust, C, C++ and Python: a flush that previously
succeeded — silently storing a mis-located geohash — now fails. There is no
changelog in this repo and `doc/` does not document geohash, so this needs a
release note.

## Open findings — not ready to merge as-is

**1. The new bound is wrong at both ends (blocking).** The QWP wire has a
sentinel NULL mode: with no null bitmap, the server decodes all-ones over
`ceil(bits/8)` bytes as NULL (`QwpGeoHashColumnCursor.java:72-76`, with
`valueSize = (precision + 7) / 8` at `:164`).

- The check now *rejects* that sentinel at 113 of 116 reachable `(dtype, bits)`
  combinations, including `GEOHASH(60b)` — so a bitmap-less column using `-1`
  for NULL, which previously ingested as SQL NULL, hard-fails the whole flush.
- Conversely the advertised maximum, `values must be in 0 .. (1<<bits)-1`, *is*
  the sentinel at the seven byte-aligned precisions (8/16/24/32/40/48/56). Worse,
  `use_bitmap = kind_supports_sparse_nulls(kind) && null_count > 0`
  (`arrow_batch.rs:3281`) means the same value lands as a real geohash when the
  batch has any null row and as NULL when it has none.
- The `bits >= src_width_bits` exemption's premise ("every value is in range")
  is contradicted by `ColumnType.java:61-68` — `GEOBYTE_MAX_BITS = 7`,
  `GEOSHORT = 15`, `GEOINT = 31`, `GEOLONG = 60`; a real geohash never occupies
  the top bit of its storage width. Two new tests
  (`geohash_filling_the_{column,source}_width_keeps_its_negative_half`) assert
  `-1` reaches the wire, so fixing this means deleting tests.

**2. Measured hot-path regression.** `check_geohash_range` skips only at
`bits >= SRC*8`, and `GeohashI64` caps at 60, so every i64 geohash column pays a
full extra pass. On a 16M-row no-null `bits=60` flush: 12.06 ms -> 20.44 ms
(**+69%**), column-only encode 3.9x slower; +35% with nulls. A branchless
OR-reduce measured at +2.31 ms instead of +8.38 (3.6x cheaper) with identical
messages and all tests passing. Separately, the arrow guard is not hoisted out of
the row loop — Int32/bits=32, where the check is a tautology, still costs +6%.

**3. Cross-surface inconsistency.** Four geohash value surfaces, three
behaviours: `bind_geohash` checks and returns `InvalidBind` (documented for
geohash range at `line_sender.h:177-179`); the two paths here check and return
`ArrowIngest` / `InvalidApiCall` with different message shapes; and
`Buffer::column_geohash` (`buffer.rs:1538`) still truncates silently — the
surface C/C++/Python expose directly, and the oracle
`system_test/arrow_fuzz_common.py` compares the Arrow path against.

**4. Error loses the column name on the Chunk path.** `encoder.rs:815` calls
`write_arrow_column_body` with a bare `?` and no `decorate_column`, so
`qwp_chunk_append_arrow_column` reports `GEOHASH column: row N ...` with no
column name — while the numpy arm ten lines below, changed by this commit, now
names its column.

## Verification

`cargo fmt --check` clean; `cargo clippy --tests` clean (the one
`useless use of vec!` warning is pre-existing in `polars.rs`);
`cargo test --lib --features=almost-all-features,arrow,polars` 2419 passed,
0 failed, 23 ignored, including all 35 geohash tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to prevent invalid GEOHASH values from being silently truncated.
  * Reports the affected column and row when a GEOHASH value exceeds its declared precision or is negative.
  * Preserves valid full-width signed GEOHASH values and correctly ignores null entries.
  * Improved error reporting for invalid GEOHASH data across Arrow and NumPy ingestion.

### DIFF
--- a/questdb-rs-ffi/src/lib.rs
+++ b/questdb-rs-ffi/src/lib.rs
@@ -5368,9 +5368,9 @@ mod tests {
         padded.extend_from_slice(&bit_len.to_be_bytes());
 
         let mut words = [0u32; 80];
-        for chunk in padded.chunks_exact(64) {
-            for (idx, word) in chunk.chunks_exact(4).enumerate() {
-                words[idx] = u32::from_be_bytes([word[0], word[1], word[2], word[3]]);
+        for chunk in padded.as_chunks::<64>().0 {
+            for (idx, word) in chunk.as_chunks::<4>().0.iter().enumerate() {
+                words[idx] = u32::from_be_bytes(*word);
             }
             for idx in 16..80 {
                 words[idx] = (words[idx - 3] ^ words[idx - 8] ^ words[idx - 14] ^ words[idx - 16])
--- a/questdb-rs/src/egress/decoder.rs
+++ b/questdb-rs/src/egress/decoder.rs
@@ -1138,7 +1138,7 @@ fn reverse_uuid_rows(buf: ColumnBuffer) -> ColumnBuffer {
         Ok(values) => values,
         Err(shared) => BytesMut::from(&shared[..]),
     };
-    for row in values.chunks_exact_mut(16) {
+    for row in values.as_chunks_mut::<16>().0 {
         row.reverse();
     }
     ColumnBuffer {
@@ -1743,13 +1743,13 @@ fn count_nulls(bitmap: &[u8], row_count: usize) -> usize {
     // bit-scan to find the *position* of a null), switch to
     // `from_le_bytes` — positions are endian-sensitive.
     let body = &bitmap[..full_bytes];
-    let mut chunks = body.chunks_exact(8);
+    let (words, tail) = body.as_chunks::<8>();
     let mut nulls: usize = 0;
-    for c in chunks.by_ref() {
-        let w = u64::from_ne_bytes(c.try_into().unwrap());
+    for c in words {
+        let w = u64::from_ne_bytes(*c);
         nulls += w.count_ones() as usize;
     }
-    for b in chunks.remainder() {
+    for b in tail {
         nulls += b.count_ones() as usize;
     }
     if tail_bits != 0 {
@@ -2309,7 +2309,7 @@ mod tests {
         });
 
         let mut expected = input;
-        for row in expected.chunks_exact_mut(16) {
+        for row in expected.as_chunks_mut::<16>().0 {
             row.reverse();
         }
         assert_eq!(reversed.values, expected);
@@ -2332,7 +2332,7 @@ mod tests {
         });
 
         let mut expected = input.clone();
-        for row in expected.chunks_exact_mut(16) {
+        for row in expected.as_chunks_mut::<16>().0 {
             row.reverse();
         }
         assert_eq!(reversed.values, expected);
--- a/questdb-rs/src/ingress/column_sender/arrow_batch.rs
+++ b/questdb-rs/src/ingress/column_sender/arrow_batch.rs
@@ -925,9 +925,14 @@ fn write_qwp_bitmap_from_arrow(out: &mut Vec<u8>, nulls: &NullBuffer) -> Result<
         let word_bytes = (full_bytes / 8) * 8;
         let (src_words, src_rem) = src_slice.split_at(word_bytes);
         let (dst_words, dst_rem) = dst_slice.split_at_mut(word_bytes);
-        for (dchunk, schunk) in dst_words.chunks_exact_mut(8).zip(src_words.chunks_exact(8)) {
-            let w = u64::from_ne_bytes(schunk.try_into().unwrap());
-            dchunk.copy_from_slice(&(!w).to_ne_bytes());
+        for (dchunk, schunk) in dst_words
+            .as_chunks_mut::<8>()
+            .0
+            .iter_mut()
+            .zip(src_words.as_chunks::<8>().0)
+        {
+            let w = u64::from_ne_bytes(*schunk);
+            *dchunk = (!w).to_ne_bytes();
         }
         for (d, &s) in dst_rem.iter_mut().zip(src_rem) {
             *d = !s;
@@ -1193,7 +1198,7 @@ fn write_uuid_be_payload(out: &mut Vec<u8>, arr: &FixedSizeBinaryArray) -> Resul
             // Bulk path: slice `value_data()` once, starting at the array's
             // own offset, instead of a bounds-checked `value(row)` per row.
             let start = arr.offset() * 16;
-            for row in arr.value_data()[start..start + bytes].chunks_exact(16) {
+            for row in arr.value_data()[start..start + bytes].as_chunks::<16>().0 {
                 out.extend_from_slice(&reverse_uuid_bytes(row));
             }
         }
--- a/questdb-rs/src/ingress/column_sender/arrow_batch.rs
+++ b/questdb-rs/src/ingress/column_sender/arrow_batch.rs
@@ -2063,6 +2063,36 @@ fn geohash_bytes_per_value(bits: u8) -> usize {
     (bits as usize).div_ceil(8)
 }
 
+/// Whether a source value fits the claimed GEOHASH precision.
+///
+/// The wire keeps the low `ceil(bits/8)` bytes of each value and a
+/// geohash's high bits are its coarse position, so a value that does not
+/// fit reaches the database as a valid geohash for somewhere else
+/// entirely. A precision that fills the source width uses every bit,
+/// including the one that reads as a sign, so its top half is stored as
+/// negative numbers and every value is in range; a narrower precision
+/// leaves the sign bit clear, so a negative value there is out of range.
+#[inline]
+fn geohash_value_fits(v: i64, bits: u8, src_width_bits: u32) -> bool {
+    if u32::from(bits) >= src_width_bits {
+        return true;
+    }
+    v >= 0 && v < (1i64 << bits)
+}
+
+#[cold]
+fn geohash_value_out_of_range(v: i64, bits: u8, row: usize) -> Error {
+    fmt!(
+        ArrowIngest,
+        "GEOHASH column: row {} holds {}, which a GEOHASH({}b) cannot carry; \
+         values must be in 0 .. {}",
+        row,
+        v,
+        bits,
+        (1i64 << bits) - 1
+    )
+}
+
 fn write_geohash_payload(out: &mut Vec<u8>, arr: &dyn Array, bits: u8) -> Result<()> {
     let elem = geohash_bytes_per_value(bits);
     let row_count = arr.len();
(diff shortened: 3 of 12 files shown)


mtopolnik:
# Review — PR #193 · level 0

`fix(ingress): hold a GEOHASH value to its claimed precision at encode time`
Base `7fdaf84e` → head `e827a4f8`.

**Submodule:** `questdb` moves `4777a4e7 → 2d9244fe`. That commit is on `origin/master`, so it's a routine upstream sync — I only looked at how this repo uses it, not at its contents.

**Panic mode:** `questdb-rs-ffi/Cargo.toml:73-77` sets `panic = "abort"` for both release and dev, so any panic in the FFI crate kills the host process and `catch_unwind` does nothing there. This PR adds no new way to panic — every new failure returns a `Result` — so there's nothing to flag on that front.

---

## Critical

### C1 — A row that used to load as NULL now kills the whole flush

`questdb-rs/src/ingress/column_sender/arrow_batch.rs:2081`, `questdb-rs/src/ingress/column_sender/numpy_wire.rs:1278`

QWP lets a client say "this row is NULL" in two ways. One is a null bitmap. The other is to write all-ones and send no bitmap at all — the server reads all-ones over `ceil(bits/8)` bytes as NULL. See `QwpGeoHashColumnCursor.java:70-76`, with `valueSize = (precision + 7) / 8` at `:164`.

The client only sends a bitmap when the column actually has nulls: `use_bitmap = kind_supports_sparse_nulls(kind) && null_count > 0` (`arrow_batch.rs:3287`). So a geohash column with no nulls is always in the all-ones mode.

This client already knows about that convention. It's spelled out on the read side at `egress/decoder.rs:810-812` ("GEOHASH NULL sentinel per spec §11.5 is `-1` sign-extended across…") and implemented as `null_sentinel::GEOHASH_FF` at `:906`.

Here's what changes. Take an Arrow `Int64` column, `questdb.geohash_bits=60`, no nulls, one row holding `-1`:

- **Before this PR:** the low 8 bytes go out as `FF…FF`, and the server stores SQL NULL.
- **After this PR:** `geohash_value_fits(-1, 60, 64)` sees `60 < 64`, checks `v >= 0`, fails — and the entire flush errors out.

The NumPy path does the same thing through `check_geohash_range`. This fires on 113 of the 116 reachable `(dtype, bits)` combinations — every one except the three where `bits` happens to equal the source width. It's reachable from Rust, C, C++ and Python.

So a row that loaded fine yesterday now takes the whole batch down, and there's no other way to express NULL on a column that has no bitmap. It's also inconsistent with itself: `-1` is refused at `bits=60` in an i64 column but accepted at `bits=32` in an i32 column.

If refusing the all-ones value is what you want, that's a defensible call — but it needs to be stated as the intent, and it needs to be resolved together with C2, because right now the two halves disagree.

### C2 — The upper limit the check allows is itself the NULL value, and the "full width" escape hatch rests on a wrong premise

`arrow_batch.rs:2081-2098`, `numpy_wire.rs:1278-1315`

These are two symptoms of one thing: the wire has no way to say "all-ones is a real value here", and the new check doesn't account for that.

**(a) The maximum it advertises arrives as NULL.**

At every precision that's a whole number of bytes (8, 16, 24, 32, 40, 48, 56), `(1 << bits) - 1` is exactly the all-ones value for `valueSize = bits/8`. Walking it through:

> Arrow `Int64`, `questdb.geohash_bits=32`, one row holding `4294967295`, no nulls
> → `geohash_value_fits(v, 32, 64)`: `32 < 64` and `0 <= v < 2^32`, so it **passes**
> → `elem = 4`, so `FF FF FF FF` goes on the wire
> → server: `readValue(addr, 4)` = `0xFFFFFFFF`, and the NULL value is `-1L >>> 32` = `0xFFFFFFFF` → the row is read as NULL.

So the new error text — *"values must be in 0 .. 4294967295"* — names a value that lands in the database as NULL rather than as a geohash. That text is new in this PR.

It gets worse: whether that value is a geohash or a NULL depends on something entirely unrelated. `use_bitmap` keys off `null_count > 0`, so the same `4294967295` becomes a real geohash if any other row in the batch is null, and becomes NULL if none are.

The fuzz test can't catch this. `system_test/arrow_fuzz_common.py:828-843` (`_cmp_geohash_sentinel`) treats storage-width all-ones as null on both the expected and actual side, and `_gen_geohash:535` sets `vs[1] = cap` under `edge` — so the fuzzer generates this exact value every time and passes anyway.

**(b) The escape hatch is based on something that isn't true.**

The doc comment at `arrow_batch.rs:2071-2079` says: *"A precision that fills the source width uses every bit, including the one that reads as a sign, so its top half is stored as negative numbers and every value is in range."*

`ColumnType.java:61-68` says otherwise:

```
GEOBYTE_MIN_BITS  = 1    GEOBYTE_MAX_BITS  = 7
GEOSHORT_MIN_BITS = 8    GEOSHORT_MAX_BITS = 15
GEOINT_MIN_BITS   = 16   GEOINT_MAX_BITS   = 31
GEOLONG_MIN_BITS  = 32   GEOLONG_MAX_BITS  = 60
```

A geohash never uses the top bit of the integer type that stores it. This repo's own `system_test/arrow_fuzz_common.py:1126-1133` maps precisions to Arrow types the same way (`bits<=7 → int8`, `<=15 → int16`, `<=31 → int32`, else `int64`).

In other words, there is no QuestDB geohash that "fills the source width". `GEOHASH(32b)` is a GEOLONG; putting it in an Arrow `Int32` is a packing choice the client makes, and `-1` in that column is the 4-byte NULL value, not a geohash.

That means the two new tests assert the bug rather than the fix:

- `arrow_batch.rs` — `geohash_filling_the_column_width_keeps_its_negative_half` asserts `-1` and `i32::MIN` at `bits=32` encode fine.
- `numpy_wire.rs` — `geohash_filling_the_source_width_keeps_its_negative_half` asserts the raw `[-1i32, i32::MIN]` bytes reach the wire untouched.

The `-1` in those tests is the same `-1` that C1 refuses one dtype over. Fixing this means deleting or inverting both tests.

**Where I'd go with it:** pick one contract and apply it everywhere. Either (i) refuse a value whose low `elem` bytes are all-ones when no bitmap is being sent, and document the range as `0 ..= (1<<bits)-2` in that mode; or (ii) always send a bitmap for geohash columns so all-ones is never ambiguous. Then drop the `bits >= src_width_bits` escape hatch — it only exists to handle a case QuestDB's type system never produces.

---

## Moderate

### M1 — Measured slowdown on the NumPy i64 path

`numpy_wire.rs:1230, 1278-1287`

`check_geohash_range` skips itself only when `bits >= SRC * 8`. But `NumpyDtype::GeohashI64` caps at 60 bits (`numpy_wire.rs:288`), so that skip never happens for i64. Every i64 geohash column pays for a second full pass over the data — which takes away the single whole-column memcpy the emit path was built around.

I measured it: 16M rows, `bits=60`, no nulls, `rustc 1.97.1 -O`, best of three after warm-up.

| | time |
|---|---|
| emit (memcpy) alone | ~3.2 ms |
| `check_geohash_range` alone | ~12.9 ms |
| both together | ~15.9 ms (**+343%** on column encode) |

The loop doesn't get vectorised. The per-row `ptr::copy_nonoverlapping` into a stack buffer plus the early `return` on a bad value both get in the way.

A branchless OR-reduce over the column measured **~2.9 ms** — 4.4× cheaper, and about the same cost as the memcpy itself:

```rust
let mut acc: u64 = 0;
for &v in slice { acc |= v as u64; }
if (acc as i64) < 0 || (acc as i64) > max { /* rescan to find the offending row */ }
```

The reduce doesn't tell you which row is bad, so keep the current loop as the rescan. It only runs when the batch was going to fail anyway.

Smaller point on the Arrow side: the `$width_bits` test sits inside the row loop, so `Int32` with `bits=32` — where `u32::from(bits) >= 32` is the same answer on every row — still pays for a branch per row. Move it above the loop.

### M2 — Four places set a geohash value, three of them behave differently, and the one C/C++/Python reaches first still truncates

The title says the value is held to its precision at encode time. It is — on two of the four paths:

| where | what it does | error code |
|---|---|---|
| `bind_geohash` (query path) | checks the range | `InvalidBind` — documented at `include/questdb/ingress/line_sender.h:177-179` |
| Arrow (this PR) | checks the range | `ArrowIngest` |
| NumPy (this PR) | checks the range | `InvalidApiCall` |
| `Buffer::column_geohash` | **truncates, no complaint** | — |

`questdb-rs/src/ingress/buffer.rs:1540` → `buffer/qwp.rs:1739` only checks that `precision_bits` is in `1..=60`. The value itself, `bits: u64`, is never looked at, and `buffer/qwp.rs:7974-7977` just cuts it down:

```rust
if let ValueRef::Geohash { bits, .. } = cell.value {
    let le = bits.to_le_bytes();
    out.extend_from_slice(&le[..bytes_per_value]);
}
```

That's the path exported as `line_sender_buffer_column_geohash` (`questdb-rs-ffi/src/lib.rs:2143`, `include/questdb/ingress/line_sender.h:1194`, `line_sender.hpp:922`, `system_test/questdb_line_sender.py:1230`). It's the first thing a plain C or C++ caller reaches, and it's what `arrow_fuzz_common.py:764` uses as the reference the Arrow path is checked against.

I'm calling this correctness rather than polish: the exact mis-placed-geohash problem this PR exists to stop is still reachable through the most direct API, and the four paths now disagree on both whether they check and which error they raise.

### M3 — The new error loses the column name on the Chunk Arrow path

`questdb-rs/src/ingress/column_sender/encoder.rs:814`

The Arrow geohash message deliberately leaves out the column name, because `decorate_column` (`arrow_batch.rs:3858`) adds it afterwards. That's why `geohash_value_wider_than_its_precision_is_refused_naming_the_row` can assert on `'g'`.

But that decoration only happens on the `RecordBatch` path (`arrow_batch.rs:4172`). The Chunk path passes the error straight through:

```rust
arrow_batch::write_arrow_column_body(out, arrow_kind, arr.as_ref(), sym_res)?;   // :814
```

That's reachable from `qwp_chunk_append_arrow_column` (`questdb-rs-ffi/src/column_sender.rs:2484`). A Python or C caller using the chunk API gets `GEOHASH column: row 1 holds …` with no column name — while the NumPy arm nine lines further down (`:823`, changed by this PR) now does name its column. So the PR fixes the naming on one arm of a `match` and leaves the sibling arm as it was.

---

## Minor

- **The PR description doesn't match the PR.** It says *"No code changes: commit `79e3ebb9` cherry-picked onto `main`."* There are four commits: the geohash fix, a merge, a `questdb` submodule bump, and `e827a4f8 "Satisfy the lints added in Rust 1.98"` — a tree-wide `chunks_exact` → `as_chunks` rewrite across `decoder.rs`, `conn.rs`, `crypto.rs`, `qwp_ws*.rs` and the FFI tests. I checked those conversions and they don't change behaviour (`count_nulls` `remainder()` → `tail`, the word/remainder split in `write_qwp_bitmap_from_arrow`, `reverse_uuid_rows`, the hex helpers), and `as_chunks` is fine against `rust-version = "1.91.1"`. The body should just say they're in here.
- **No release note.** The body already flags this. A flush that worked before now fails, across four host languages, so it needs a note wherever release notes live.
- The commit title `Satisfy the lints added in Rust 1.98` is fine — plain English, under 50 chars.

---

## Coverage gaps

**Critical**

- **Nothing tests the case in C2(a), which is the exact problem this PR is about.** I searched `grep -rn "4294967295\|0xFFFFFFFF\|sentinel" questdb-rs/src/ingress/column_sender/{arrow_batch,numpy_wire}.rs` — no hits on the ingress side. There's no test checking that the largest value the new bound allows comes back as a geohash instead of as NULL. A test that encodes `Int64` / `bits=32` / value `(1<<32)-1` / no nulls and decodes the frame would fail today and would have caught this. `system_test/arrow_fuzz_common.py:828` can't stand in for it — its comparator calls that value null on both sides.

**Moderate**

- **Nothing pins down the cross-surface contract.** `grep -rn "column_geohash" questdb-rs/src/ingress/buffer/qwp.rs` finds tests for the precision range, pinning, and reset — none for the value. So there's no record of whether the row API is meant to truncate or to check, and M2 can drift either way without anyone noticing.
- **No C++ or Python test for the new error.** `cpp_test/` and `system_test/` are untouched. The new failure is reachable from the FFI entry point (`qwp_chunk_append_arrow_column`), the C++ wrapper, and the Python ctypes shim — and M3 is exactly the kind of thing a test from one of those would have caught.

---

## Things I checked and dropped

- **"A failure partway through a column leaves a half-written frame in the buffer."** No — `encoder.rs:249-252` truncates `out` back to `frame_start` and rolls the symbol dict back on any error, and the `RecordBatch` path does the same via `rollback_on_err` (`arrow_batch.rs:4159-4173`).
- **"`(1i64 << bits) - 1` overflows at `bits = 63`."** Can't happen. `geohash_value_out_of_range` only runs after `geohash_value_fits` returned false, which requires `bits < src_width_bits`, and `check_geohash_range` returns early at `bits >= SRC*8`. The largest shift is 59 for i64 and 31 for i32, and `NumpyDtype::validate` caps `GeohashI64` at 60 (`numpy_wire.rs:288`).
- **"`as_chunks` breaks the MSRV."** It doesn't — stable since 1.88, and `rust-version` is `1.91.1` in both crates.
- **"`check_geohash_range` walks a strided NumPy view incorrectly."** No — `chunk.rs:1367` sets `src_stride = dtype.source_elem_size()?`, always the element size, and `chunk.rs:1983-1986` checks it against `expected_stride`.
- **"`panic = "abort"` turns the new failure into a process abort."** No — every new failure returns a `Result`. No `unwrap`, `expect`, or indexing was added on caller-supplied values.

---

## Summary

**Request changes.**

- **Tests pass.** `cargo test --manifest-path questdb-rs/Cargo.toml --lib --features=almost-all-features,arrow,polars geohash` → **35 passed, 0 failed**. `cargo fmt -- --check` clean. `cargo clippy --tests` has one warning, `useless use of vec!` at `questdb-rs/src/ingress/polars.rs:679` — already there before this PR, untouched by it.
- **Coverage gaps: 3** (1 critical, 2 moderate). The critical one blocks alongside C1 and C2.
- **Submodule:** `questdb` is a routine upstream sync.
- **11 findings kept, 5 dropped after checking the source.**
- **6 findings inside the diff, 2 outside it** (M2 at `buffer/qwp.rs:7974`, M3 at `encoder.rs:814`).

The change goes after a real bug. The encoders do cut a geohash down into a perfectly valid geohash for somewhere else, and checking the value right where the truncation happens is the right place to put the check.

But the limit it enforces doesn't line up with the wire it's enforcing against. At byte-aligned precisions the maximum it lets through comes back as NULL, and at the three precisions where it turns itself off entirely, the reasoning is contradicted by `ColumnType.java` — with two new tests written to keep that behaviour. Separately, a value that loaded as NULL yesterday now takes down the whole flush on four host languages with nothing to replace it, and the NumPy i64 path pays roughly 4× on column encode for a check that a branchless reduce does at memcpy speed.

The body already flags most of this and says it isn't ready to merge. I confirmed all four of its open findings against the source, and I'd add M3's reachability from `qwp_chunk_append_arrow_column`, plus an independent measurement of the slowdown (+343% on column encode; the branchless version is ~2.9 ms against ~12.9 ms).

PR #195, opened 2026-08-27
fix(ingress): preserve maximum GEOHASH values and validate Arrow FFI inputs

mtopolnik:
Tandem PRs:

[py-questdb-client #140](https://github.com/questdb/py-questdb-client/pull/140)
[docs #516](https://github.com/questdb/documentation/pull/516)

## Summary

Fixes valid byte-aligned GEOHASH maxima being encoded as `NULL` and validates Arrow C Data structures before passing them to `arrow-rs`. It also corrects delivery reporting at the Rust Polars DataFrame boundary: a later validation failure must not hide an earlier published prefix.

## GEOHASH encoding

For byte-aligned precisions, the largest valid GEOHASH value is an all-ones bit pattern, which also represents `NULL` when the column has no validity bitmap. These columns now always include a bitmap. The encoding and size calculations agree across UDP, WebSocket, Arrow and NumPy ingestion, including frame-size estimation and row-group planning.

Bulk GEOHASH input remains permissive: values are raw bit patterns, and the encoder writes the low `ceil(precision / 8)` bytes. The client checks precision, carrier width, column consistency and input shape; applications remain responsible for checking that values fit the declared precision.

## Arrow FFI validation and ownership

The guarded Arrow FFI entry points validate schema/array agreement, child counts, column and node limits, unsupported Struct columns, metadata bounds, buffer layouts, offsets and slice lengths before materialization.

The unpublished FFI artifact and its Arrow-family dependencies are pinned to Arrow 59.0.0. The published `questdb-rs` crate retains its `>=58, <60` compatibility range. This PR includes the Arrow work developed in [#196](https://github.com/questdb/c-questdb-client/pull/196) and [#197](https://github.com/questdb/c-questdb-client/pull/197).

Supported structural errors return client errors. Producers must still supply valid, stable, correctly sized allocations and NUL-terminated strings where the Arrow ABI requires them. Checks cannot establish the validity of arbitrary pointers.

Failures detected before import retain the producer's `array->release` callback. Once import begins, ownership may have been consumed. Callers must inspect `array->release` on failure; error codes alone do not determine ownership. The existing delivery-based Arrow re-export behavior is preserved.

## Public API and delivery scope

[...]

PR #197, opened 2026-08-28
fix(arrow): model pinned FFI conversion layouts

mtopolnik:
## Summary

- encode a shared, exhaustive Arrow 59 schema/array model with exact child arity and fixed/view buffer layouts
- enforce fan-out budgets before pointer traversal and signed/i686-safe slice arithmetic
- add raw-versus-guarded differential fixtures for the Arrow 59 Struct double-slice and FixedSizeList boundaries
- pin only the unpublished questdb-rs-ffi artifact to Arrow 59.0.0 and enforce its lock graph in Cargo, wheel, and CI entry points

## Dependency

This follow-up is based on Stage 1 PR #196 at commit 61245840262ba6c687e72e67393ad2cde7740e83. PR #196 is itself based on fix/geohash-value-range / PR #195 at 33ae96e7. Retargeting and landing remain gated on that dependency chain.

## Safety boundary

The guard covers addressable Arrow C Data Interface structure, metadata, arity, buffer layout, and modeled Arrow 59 slice invariants before import. Producer-owned pointer arrays, allocations, and C strings must still be valid, correctly sized, stable, and terminated because the ABI does not expose their allocation bounds.

## Cargo invocation audit

- ci/run_all_tests.py runs the single lock guard first, generates a temporary exact-59 questdb-rs graph, verifies it with cargo metadata, and uses --locked for Arrow tests and both FFI test invocations
- ci/run_tests_pipeline.yaml runs the guard before the direct FFI MSRV check and Clippy invocations; both use --locked
- Python setup.py supplies --locked only to questdb-rs-ffi (the rpyutils build remains unchanged)
- the Python repository's ci/run_cibuildwheel.py wrapper runs the same guard before every wheel build

## Validation

- cargo fmt --manifest-path questdb-rs-ffi/Cargo.toml --check
- cargo test --manifest-path questdb-rs-ffi/Cargo.toml --locked: 88 passed
- cargo test --manifest-path questdb-rs-ffi/Cargo.toml --all-features --locked: 141 passed
- cargo clippy --manifest-path questdb-rs-ffi/Cargo.toml --all-targets --all-features --locked -- -D warnings
- questdb-rs Arrow test graph at exactly Arrow 59.0.0: 1,942 passed, 23 ignored, plus all integration and doctest groups
- python3 ci/check_arrow_ffi_lock.py: 11 Arrow packages at 59.0.0
- python3 ci/check_docs.py: 12 files, 85 local links

The upstream arrow-rs report URL and its matching source comment remain a tracked finishing item; posting the prepared external issue requires explicit disclosure approval.

### DIFF
--- a/ci/check_arrow_ffi_lock.py
+++ b/ci/check_arrow_ffi_lock.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+"""Verify the Arrow implementation pinned into questdb-rs-ffi artifacts."""
+
+import pathlib
+import tomllib
+
+
+EXPECTED_ARROW_VERSION = '59.0.0'
+EXPECTED_ARROW_REQUIREMENT = f'={EXPECTED_ARROW_VERSION}'
+REPO_ROOT = pathlib.Path(__file__).resolve().parents[1]
+FFI_ROOT = REPO_ROOT / 'questdb-rs-ffi'
+
+
+def fail(message):
+    raise SystemExit(f'Arrow FFI lock guard failed: {message}')
+
+
+def main():
+    with (FFI_ROOT / 'Cargo.toml').open('rb') as manifest_file:
+        manifest = tomllib.load(manifest_file)
+    arrow_dependency = manifest.get('dependencies', {}).get('arrow')
+    if not isinstance(arrow_dependency, dict):
+        fail('questdb-rs-ffi must declare arrow as a dependency table')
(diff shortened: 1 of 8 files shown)


mtopolnik:
Superseded by the clarified review-fix landing structure: these commits have been fast-forwarded directly onto support PR #195. The branch is retained; no work was deleted.

PR #198, opened 2026-08-28
test(arrow): harden FFI regression boundaries

mtopolnik:
## Summary

Completes the Stage 4 regression and raw-read audit follow-up for the pinned Arrow 59 FFI model from #197.

This PR is intentionally based on `fix/arrow-ffi-hardening-model` at `30d462e1cdf639466ddbc558a6e44919ff5b8016`. Its single commit is `82096e6b5fcbd3c60b7ceb44ee658af546ce332e`; the dependent base must land first.

## Coverage added

- exercises the real record-batch and column entry points for nested Struct rejection at root offsets 0 and 1, a narrowed zero-row view, a three-level Struct, and supported/short FixedSizeList chains;
- imports and materializes every accepted Arrow 59 classifier arm and physical-layout family, including temporal-unit boundaries, fixed-size binary widths, views, decimals, lists, and every supported Dictionary key/value storage combination;
- makes root-column, total-node fan-out, non-root arity, exact/bounded buffer layout, row/offset, i686-representative arithmetic, and metadata-budget diagnostics deterministic and distinct;
- expands the fully allocated Arrow 59 differential fixtures without relying on undersized allocations or poison pointers that can be reached;
- records the residual producer-memory contract beside every raw Arrow-path C-string, metadata, child-pointer, buffer-pointer, and variadic-length read;
- records the complete downstream `from_ffi` / `ArrayData::new_unchecked` / validation / materialization sequence reached after preflight.

## Safety boundary

The guarded Python/C-ABI path rejects addressable malformed structural values before the modeled Arrow 59 panic paths. Producer-owned pointers and their allocations must still be valid, correctly sized, stable, and NUL-terminated where required until release; the Arrow ABI does not expose allocation lengths that would let the consumer prove otherwise.

The arrow-rs nested-Struct report is deliberately held pending explicit disclosure approval. The local offset-0/offset-1 differential test remains the pinned regression gate.

## Validation

- `cargo fmt --check` for `questdb-rs-ffi` and `questdb-rs`
- `python3 ci/check_arrow_ffi_lock.py` — 11 Arrow-family packages at 59.0.0
- `cargo test --locked --manifest-path questdb-rs-ffi/Cargo.toml` — 88 passed
- `cargo test --locked --all-features --manifest-path questdb-rs-ffi/Cargo.toml` — 147 passed
- strict all-target/all-feature `cargo clippy --locked` — passed
- complete Python suite in the dependent superproject — 1,014 passed, 28 skipped

The existing C/C++ ASan+UBSan job does not execute these Rust FFI unit fixtures, and there is no Rust sanitizer job for them. All native differential fixtures are fully allocated, and the complete deterministic subprocess suite runs in the dependent Python PR.

## Separately tracked Python follow-ups

- questdb/py-questdb-client#147
- questdb/py-questdb-client#148
- questdb/py-questdb-client#149

### DIFF
--- a/questdb-rs-ffi/src/lib.rs
+++ b/questdb-rs-ffi/src/lib.rs
@@ -4169,6 +4169,8 @@ unsafe fn arrow_format_str<'a>(
                 "Arrow schema {path}: format pointer is NULL"
             )));
         }
+        // RAW-READ AUDIT: NULL is rejected above; termination and allocation
+        // remain producer-memory obligations because the ABI exposes no size.
         let cstr = std::ffi::CStr::from_ptr(p);
         cstr.to_str().map_err(|_| {
             arrow_ingest_err(format!("Arrow schema {path}: format string is not UTF-8"))
@@ -4176,7 +4178,7 @@ unsafe fn arrow_format_str<'a>(
     }
 }
 
-// `FFI_ArrowSchema::name()` in arrow-schema-58.x calls `.expect("non-utf8
+// `FFI_ArrowSchema::name()` in arrow-schema 59 calls `.expect("non-utf8
 // as name")` on every import, and `TryFrom<&FFI_ArrowSchema> for Field`
 // invokes it unconditionally. Under `panic = "abort"` an invalid byte in
 // `name` from an Arrow producer aborts the host. NULL is allowed (treated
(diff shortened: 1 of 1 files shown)


mtopolnik:
Superseded by the clarified review-fix landing structure: these commits have been fast-forwarded directly onto support PR #195. The branch is retained; no work was deleted.

PR #203, opened 2026-09-07
fix(qwp): allow symbols after columns

jerrinot:
QWP/UDP and QWP/WebSocket row buffers now accept symbol and non-symbol columns in any order before the row timestamp. Previously, adding a symbol after another column failed with InvalidApiCall.

ILP continues to require symbols before fields.

## Summary by CodeRabbit

- **New Features**
  - QWP rows now support interleaving symbol and non-symbol columns before the timestamp across UDP and WebSocket connections.
  - ILP rejects symbols added after non-symbol columns.

- **Bug Fixes**
  - Validation messages now include `symbol` among allowed operations.
  - Rejected ILP symbol calls do not modify the in-progress row.

- **Documentation**
  - Clarified column-ordering requirements for ILP and QWP across supported language APIs.
  - Expanded guidance for sequential row construction and buffering after endpoint failures.


jerrinot:
## Review: `fix(qwp): allow symbols after columns`

Base `cd59ad05` → head `a73d7b48`. Adversarial multi-pass review; all behavioural claims below were verified by executing code, with base-vs-head control runs on identical triggers.

### Test gate — green

`almost-all-features` (2143 lib tests), ILP-only, QWP/WS-only, QWP/UDP-only, `questdb-rs-ffi --features arrow`, `cargo clippy --tests` (both crates, no `-D warnings`), `cargo fmt --check`, and the CMake `cpp_test/test_line_sender` suite (95 cases / 1892 assertions) including both new cases. All pass.

Mutation check: reverting both call sites fails `qwp_udp_allows_symbols_after_non_symbol_columns` and `qwp_ws_columnar_allows_symbols_after_non_symbol_columns`, while `unordered_symbols_may_follow_columns` still passes — exactly as the comment at `tests/sender.rs:336-339` claims. One flaky failure under parallel load (`qwp_ws_background_role_reject_recycles_wire_and_drains_orphan`, connection reset), 5/5 in isolation, unrelated.

### The change itself is correct

Three independent fuzz runs over the encoders: 2000 randomised buffers with randomised symbol offsets showing exact `Buffer::len()` == wire bytes; a 300-seed UDP size-hint test asserting exactness for every row prefix; a 300-seed WS encode→decode round-trip against a shadow model — each with a base control run that fails for the expected reason. Multi-row interleaving, null-padding, late-created symbol columns and interleaved rollback all encode byte-identically between the row-log and columnar encoders. No encoder, size-hint, state-machine or bounds defect exists.

**Two premises settled:**

- **Server acceptance is not a risk.** The "symbols are a contiguous prefix" property *never held* — both QWP buffers order columns by first touch across the whole segment, not per row (`qwp.rs:3543`, `qwp.rs:6934-6936`). At base, `table("t").column_i64("a",1).at_now(); table("t").symbol("s","x").column_i64("a",2).at_now()` already emits schema `[a:LONG, s:SYMBOL]` in one datagram — verified by execution on `cd59ad05`. The chunk (`qwp_chunk_*`) and Arrow paths already place symbol columns at any index. The PR creates no new wire shape. This is the strongest support for its premise.
- **Hot-path cost is zero.** In the built `.so` the inlined check is `movzbl / lea -0x1 / cmp $0x3 / jae <cold>`; base emitted the same three instructions with `cmp $0x2`. LLVM folded the `||` into one unsigned range test. A mask-based `unordered_allowed_ops()` refactor compiles to code LLVM's ICF aliases with the current form — so don't do it on performance grounds.

**Safety:** no newly-reachable panic, abort or unsoundness. `qwp.rs` and `buffer.rs` contain zero `unsafe`. The load-bearing invariant — the `touched_* <= old_*` guards backing unchecked `usize` subtractions in `RowGroupPlanner::add_row` (`qwp.rs:7062-7068`, `7079`, `7103`) — holds because the relaxation reorders a row's entries without changing the set of names in it.

---

## Critical

None.

## Moderate

### M1. QWP/WS silently discards a repeated symbol value written after a column; QWP/UDP errors

`qwp.rs:3586-3588` (out-of-diff). Trigger: `table("t").symbol("s","a").column_i64("q",1).symbol("s","b")`

- head WS: returns `Ok`, row commits, dictionary holds `"a"` only. `"b"` is gone with no signal.
- head UDP: `InvalidApiCall: column 's' already set for current row`; row preserved.
- base, both: the ordering error. Uniform, safe, nothing written.

The `if duplicate { return Ok(()); }` mechanism is byte-identical at base and already reachable via two adjacent `symbol()` calls — but for *this* trigger the PR converts a uniform error into silent value loss on the default transport. The motivation for the relaxation is generic / map-driven emitters, which are precisely the emitters that can emit a name twice.

### M2. QWP/WS destroys the in-progress row when a symbol fails after a column, and `check_can_flush()` still returns `Ok`

`qwp.rs:3531-3541`, `3580-3584`, `3641` (out-of-diff). Trigger: `table("t").column_i64("q",1).symbol("bad?name","v")`

- head WS: `Err(InvalidName)` **and** `rollback_current_row()` — `row_count=0`, a following `at_now()` fails with "Bad call to `at`". Because `check_can_flush()` returns `Ok`, a caller that logs the error and continues flushes a buffer silently missing the row.
- head UDP: `Err(InvalidName)`, row preserved, `at_now()` succeeds.
- base, both: row preserved.

This PR's own new test asserts exactly the violated invariant for ILP (`tests/sender.rs:344-356`: *"The rejected call wrote nothing: the row is still completable"*). There is no QWP equivalent.

**M1 and M2 sit next to a warning this branch wrote and then dropped.** `24815dbb` documented both divergences on five surfaces — `doc/CONSIDERATIONS.md`, `line_sender.h`, `line_sender.hpp`, `questdb-rs-ffi/src/lib.rs` and `questdb-rs/src/ingress/buffer.rs` ("*a later value may fail, be silently ignored, or be encoded as a separate column… Type-change errors… may discard the in-progress row*") — and `a73d7b48` removed it from all five as "duplicate-value guidance" while keeping the change that makes them more reachable. To be precise about the blast radius: nothing regresses against `main`, which never carried that text. The point is only that the warning was written, judged accurate, and then dropped from the same PR that widens what it warns about. Restore it or open a follow-up.

### M3. `ErrorCode::InvalidApiCall`'s doc example is now false on three public surfaces

`questdb-rs/src/error.rs:51`, `questdb-rs-ffi/src/lib.rs:264`, `include/questdb/ingress/line_sender.h:90` — all read *"Called methods in the wrong order. E.g. `symbol` after `column`."* A QWP user looks up the code, concludes the row will be rejected, and writes an unnecessary two-pass fill — the opposite of what this PR enables. Nothing guards it: the two enum-mirror tests (`lib.rs:5011`, `:5050`) compare names and discriminants only, never doc text. `84c79f4c` fixed all three ("…on an ILP buffer") and `24815dbb` reverted them. Three lines, zero test churn.

### M4. The flowchart this PR updated reaches no reader, and the prose beside it contradicts it

Verified by building rustdoc and grepping the output. `seq.dot` gained `column -> symbol [label="QWP only"]`. Its only in-repo reference is `ilp.rs:212`, inside a doc comment on `ilp::Buffer` in the **private** `mod ilp` (`buffer.rs:25`, re-exported `pub(crate)` at `:36`) — it renders only in `target/doc/src/…/ilp.rs.html`, the source viewer, never an item page. Meanwhile `mod.md:387` (a line this PR edited) says *"Refer to the [`Buffer`] doc for the full rules and a flowchart"*, and the public `Buffer` doc (`buffer.rs:389-393`) is three lines with neither. Six lines above the image, `ilp.rs:206` still states *"Symbols must appear before columns."* unqualified. The "# Sequential Coupling" block needs to move onto the public `Buffer`, or the only artifact edit in this PR is dead weight.

### M5. ILP's rejection message steers the user into a different column type and names no protocol

Current output: ``State error: Bad call to `symbol`, should have called `column` or `at` instead.`` Followed literally → `column_str` → per the type table at `ilp.rs:218-222` that is STRING/VARCHAR, not SYMBOL, and the server materialises it permanently on first write.

House style for protocol-dependent calls is the opposite: `column_i8` errors with `"column_i8 requires a QWP transport (ws:: or udp::)"` — that phrasing appears 14× in `buffer.rs` and is mirrored in the C header (`line_sender.h:955-957`). This bites now because default `sync-sender` bundles TCP + HTTP + QWP/WS and `Sender::new_buffer()` (`sender.rs:285`) picks the flavour from the handler, so a conf-string change from `ws::` to `tcp::` flips the rule with no code change. Nothing pins this string — zero test churn. `84c79f4c` added exactly this and `24815dbb` reverted it deliberately: that commit's message is "allow symbols after columns **without changing ILP error guidance**". That is a defensible scoping call, and I'd still push back on it — the PR is what creates the asymmetry, so the message that ILP users now hit is this PR's to own, not a pre-existing wart it happened to touch.

### M6. The QWP/WebSocket test asserts almost nothing about the encoded frame

`qwp.rs:10172`'s only assertion is `assert_eq!(entries, vec![b"ETH-USD", b"XNAS"])` from `ws_delta_entries`, which (`qwp.rs:8069-8089`) parses the header plus the symbol-dictionary delta and returns — it never reads the table block. The test passes if the encoder hoists both symbols to the front (the exact failure its name guards), reorders columns, mis-attaches symbol ids, or omits `qty`/`active` entirely. It is also an internal buffer unit test that never touches public `Buffer`/`Sender`.

Worst placement: QWP/WS is the default transport (`sync-sender` → `sync-sender-qwp-ws`); the one order-and-value assertion in this PR (`tests/qwp.rs:175`) covers opt-in, best-effort UDP. `24815dbb` had a positional `ws_single_row_columns` decoder for exactly this and `a73d7b48` deleted it. Model to copy: `tests/qwp_ws.rs:1926` `qwp_ws_round_trip_minimal_message`, ~25 lines, no new infrastructure.

### M7. The proptests never generate the newly-legal row shape

`apply_row` (`qwp.rs:8778`) writes `buf.symbol("sym", …)` before every column, so all five proptests (`qwp.rs:9201`, `9219`, `9240`, `9268`, `9288`) see only symbols-first rows — including the ones checking size-hint exactness, datagram splitting and marker rewind, which are the invariants most sensitive to entry ordering. `84c79f4c` added a generated `symbol_offset`; the revert removed it. The fuzzing above found no defect, so this is a removed guard rather than a hidden bug — but it is the highest-leverage single fix, since one generated position fuzzes encode/decode *and* splitting for free.

### M8. No test sends an interleaved row to a real QuestDB server

Every added test validates against the client's own decoder. The real-server harness is structurally blind: `system_test/qwp_ws_fuzz.py:1188` emits all symbols in one loop then all columns in a second, so `column_reordering_factor` shuffles only within each group; the ~30 `test_*_qwp_udp` cases are all symbol-first; and the QWP/WS system tests use no symbol at all (`_create_qwp_ws_table`, `test.py:1618`, creates `id`/`val`/`ts`).

Graded Moderate rather than Critical: since the interleaved wire shape is already emittable at base, the server is already receiving it today, so this is regression protection rather than an unverified assumption. Still worth one test modelled on `test.py:3473`, which runs in CI.

## Minor

- **`allows_symbol()` (`op_state.rs:189`) now contradicts `check_unordered` and carries no doc.** It returns `false` in `ColumnWritten` where `check_unordered(Op::Symbol)` returns `Ok`. Its sole non-test caller is the ILP field separator (`ilp.rs:611`), where `false` is correct — but nothing says so, and the next QWP caller reaching for the obvious name silently reinstates the ILP rule. Two lines of doc, or `84c79f4c`'s `ilp_symbol_section_is_open` rename.
- **The relaxed rule is stated twice**: `op_state.rs:128` encodes the op set, `op_state.rs:94-99` hand-copies the advice literal from `SymbolWritten`'s arm. (There is *no* "new `OpCase` variant causes drift" risk — `next_op_descr` matches all five variants with no catch-all, so a new variant is a compile error, and `_ =>` delegation inherits the strict text.)
- **The QWP mid-row error string changed** (the advice tail went from "should have called ‘column’ or ‘at’" to "should have called ‘symbol’, ‘column’ or ‘at’"). In-tree consumers are updated; out-of-tree string matchers break — notably py-questdb-client, which links this same cdylib. No CHANGELOG in this repo, so it needs a release note.
- **`qwp.rs:11141` `qwp_state_machine_matches_ilp_shape`** — QWP deliberately no longer matches the ILP shape; it passes only because its assertions stop before the changed transition.
- **`tests/qwp.rs:1858` and `:1866` now assert byte-identical strings**, so `qwp_buffer_check_can_flush_tracks_public_state_machine` no longer distinguishes the two states it walks.
- **No assertion anywhere for "Bad call to `table`"** — a repo-wide grep for that message returns zero hits, and this PR changed that message's text. One assertion closes it.
- **`doc/CONSIDERATIONS.md:116-117`** appends the row-ordering sentence to a paragraph whose subject is the *columnar* QWP API (Arrow/Polars, no `symbol()`/`at()`).
- **`Buffer::new_qwp()` (`buffer.rs:440`) has no doc comment** while `new_qwp_ws()` does; `symbol_opt` (`:797-800`) doesn't restate the protocol qualifier where its `column_*_opt` siblings do.
- **No issue link in the PR description** despite the `bug` label.

## Considered and dismissed

- **Case-folding divergence** — `column_i64("Q",1).symbol("q","v")`: UDP accepts and emits `Q`+`q` as two columns (exact-bytes comparison at `qwp.rs:1309`), WS folds and rejects (`:3927`). Real, but reachable at base symbol-first. Pre-existing; follow-up material.
- **Server may reject non-leading symbol columns** — settled above; base already emits the shape.
- **`check_unordered` branch cost / `column_access_cursor` mispredicts** — zero added instructions; cursor hit rate depends only on per-row call-order consistency, 100% under both rules.
- **`Send`/`Sync`, symbol-dictionary delta, SFA replay, Java golden bytes, drainer paths** — all clean. `OpState` is `Copy` and taken by value; dictionary ids are position-indexed and self-consistent; replay stores opaque frames with inline schemas and never re-encodes; the wire format has no symbols-first field.
- **Latent partial-append rollback** (`qwp.rs:3586-3596` vs `:4067`) — requires `u32::MAX` overflow.
- **C++ test resolves cells by name and is order-blind** — redundant; `tests/qwp.rs:175` pins UDP order.
- **ABI** — filtering the diff for `include/` and `questdb-rs-ffi/` down to non-comment lines returns nothing. C, C++ and Rust FFI doc text are byte-identical; headers are hand-maintained by design.

---

## Summary

**Approve with comments.** No Critical findings and no Critical coverage gap. The design call is right — this makes the row API consistent with the chunk and Arrow APIs on the same transport — the implementation is provably free and correct, the cfg gating is clean across all four feature combinations, and the whole test matrix plus the C++ suite is green.

The weakness isn't the code, it's what the last two commits trimmed on the way to it — and they trimmed different things, so it's worth separating them. `24815dbb` scoped out the ILP-side work on purpose ("without changing ILP error guidance"), which is where the `InvalidApiCall` doc fix (M3) and the protocol-naming error text (M5) went. `a73d7b48` then removed the positional WS decoder and the C++ schema-order assertions (M6), the proptest symbol offset (M7), and the five-surface divergence guidance (M1, M2). Every one of them is a small restore from `84c79f4c` or `24815dbb` rather than new work.

M1 and M2 are the ones I'd not merge without at least a follow-up issue — silent value loss and silent row loss on the default transport, in a client library, deserve to be written down somewhere even if they aren't fixed here.

13 findings admitted (0 Critical, 8 Moderate, 5 Minor groups); 7 candidates omitted after verification. 5 in-diff, 11 out-of-diff. No submodule pointer changed.

---

*Edited to correct commit attribution: the `InvalidApiCall` doc fix (M3) and the ILP error text (M5) were reverted by `24815dbb`, not `a73d7b48`; and the divergence guidance landed on five surfaces, not four.*

jerrinot:
## Review summary

**Approve with non-blocking test/documentation suggestions.** No confirmed blocking regression found.

The QWP state-machine change permits symbols after non-symbol columns while preserving ILP ordering restrictions.

Suggested follow-ups:
- Strengthen the new WebSocket regression test by decoding the table payload and asserting column names, types, and values; it currently checks only the symbol dictionary.
- Qualify the `InvalidApiCall` documentation example (`symbol` after `column`) as applying to ILP buffers.

Clarification on existing behavior: WebSocket already uses first-value-wins for duplicate columns and rolls back the current row on certain validation failures. This PR makes those existing behaviors reachable after a non-symbol column; it does not introduce either policy. The WS/UDP inconsistencies merit separate follow-up, not a blocking finding here. The missing flowchart on public `Buffer` documentation is also pre-existing.

Verification performed on head `a73d7b48`: Rust library suite (2,143 passed, 23 ignored), targeted symbol tests across ILP/TCP, QWP/WS, and QWP/UDP feature configurations, formatting checks and clippy for both Rust crates, and public-API probes of duplicate handling and row rollback.

Review limitation: independent subagents were unavailable, so the requested level-3 independent-agent review was not completed. Findings were checked directly against source and executable probes.

jerrinot:
### Critical

None.

### Moderate

None.

### Minor

#### m1. ILP’s late-symbol error message is now misleading

- **File:** `questdb-rs/src/ingress/buffer/op_state.rs:83-88`
- **Type:** API ergonomics / error-message regression

ILP still reports:

```text
State error: Bad call to `symbol`, should have called `column` or `at` instead.
```

After this PR, `symbol` after a non-symbol column is protocol-specific: QWP accepts it, ILP rejects it. The error should say that ILP requires symbols before non-symbol columns, instead of suggesting another `column_*` call.

Suggested fix: special-case `Op::Symbol` after `ColumnWritten` in the strict/ILP path and make the message protocol-specific.

#### m2. No direct pure-C ABI regression for QWP symbol-after-column

- **Files:** `cpp_test/test_line_sender.cpp`, `questdb-rs-ffi/src/lib.rs:1590-1599`

The C++ wrapper and Python ctypes paths are covered, but there is no direct C-style test using `line_sender_buffer_column_*` followed by `line_sender_buffer_symbol` on a QWP buffer.

Suggested fix: add a C API test mirroring the new C++ UDP case.

#### m3. No Rust public QWP/WebSocket mock-server regression

- **File:** `questdb-rs/src/tests/qwp_ws.rs`

The internal encoder test and Python system test cover QWP/WS interleaving, but the Rust public `Sender`/`Buffer` mock-server path lacks a targeted `column -> symbol` test.

Suggested fix: add a small `qwp_ws.rs` mock-server test beside the existing QWP/WS round-trip case.

### Downgraded (false positives)

- **QWP/WebSocket drops the in-progress row after late-symbol validation/type errors** — dismissed as blocking: the rollback policy is pre-existing; this PR only makes one additional call order reach it. Track/document separately if desired.
- **QWP encoders assume symbols are a prefix** — dismissed: QWP/WS lookup falls back to column map, and tests decode interleaved column order/values.
- **ILP now accepts symbol after column** — dismissed: ILP still uses strict `OpState::check()`.
- **FFI ABI changed** — dismissed: `line_sender_buffer_symbol` signature is unchanged across Rust/C/C++/Python.
- **New panic/abort path from `check_unordered`** — dismissed: the new check is enum branching plus cold error formatting; no unsafe/indexing/unwrap.

### Summary

**Verdict: approve with non-blocking comments.**

No blocking correctness, FFI-safety, panic, ABI, or encoder regression found. The remaining issues are error-message clarity and extra coverage suggestions.

3 minor findings verified, 5 false positives/downgrades removed.

bluestreak01 (approved):
Reviewing PR #203 at level 3, head 8abfc0de398e92f1f4726383fbf5bbfd555e6a7d, base cd59ad0586330730e2e0eb65239afb45ad17ca39.

### Findings
None: **0 Critical, 0 Moderate, 0 Minor**.  
Admitted split: 0 in-diff / 0 out-of-diff breakages.

### Coverage map
**Test gate: passes. Admitted coverage gaps: 0.**

- 216 targeted Rust tests passed, plus isolated TCP/UDP/WebSocket feature checks.
- C++ interleaving test passed against the abort-mode FFI build; identical input produced the expected ordering error at base.
- Reverting the production changes made the corresponding regression tests fail.
- Rust formatting and clippy checks completed without newly introduced warnings.
- [TestVsQuestDBMaster](https://dev.azure.com/questdb/questdb/_build/results?buildId=268647) logs show the changed live round-trip and failover tests passing on the PR merge ref containing this head. Live tests were not rerun locally.

### Summary
**Verdict: approve.** Correctness and test gates pass.

QWP ordering is relaxed; ILP remains strict. Core panic mode defaults to unwind; FFI dev/release use abort. Overflow checks are enabled in dev/test and disabled in release.

Submodules: no pointer changes. Snapshot revalidated; source trees unchanged.

Evidence: `/tmp/pr203-review-51Gtn0/`.

